PR #2, opened 2020-01-23
Complete higher order function challenges #ticketnumber

drewcpete:
Adds 8 new functions using higher order functions and has a single test for each function.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -126,18 +126,18 @@ export const animalTypeCountFilter = (petArray) => {
     }, {})
 }
 
-console.log(tests.testAnimalCount() + " testing for animal type count")
+console.log(tests.testAnimalCount() + " for animal type count")
 
-console.log(tests.testAnimalFightByName() + " testing animal fight by name")
+console.log(tests.testAnimalFightByName() + " animal fight by name")
 
 console.log(tests.testDoubleDanger() + " double danger test");
 
-console.log(tests.testFindDanger() + " testing find most danger")
+console.log(tests.testFindDanger() + " find most danger")
 
-console.log(tests.testFindLeastDanger() + " testing find least danger")
+console.log(tests.testFindLeastDanger() + " find least danger")
 
-console.log(tests.testFindDangerAndType() + " test find danger and type");
+console.log(tests.testFindDangerAndType() + " find danger and type");
 
-console.log(tests.testAlphabetizePets() + " test alphabetize pet list")
+console.log(tests.testAlphabetizePets() + " alphabetize pet list")
 
-console.log(tests.testAnimalFight() + " testing animal fight")
+console.log(tests.testAnimalFight() + " animal fight")
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,7 @@
-const pets = [
+import * as tests from "./testing"
+
+
+export const pets = [
     {
         name: "fido",
         type: "dog",
@@ -20,7 +23,7 @@ const pets = [
         dangerLevel: 5,
     },
     {
-        name: "CatStevns",
+        name: "CatStevens",
         type: "cat",
         dangerLevel: 7,
     },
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1,4 +1,7 @@
-const pets = [
+import * as tests from "./testing"
+
+
```
frankmeza: rm this extra line

src/index.ts
```diff
@@ -36,25 +39,105 @@ const pets = [
...
 
+
+// New Array
+// write a function that returns an alphabetized list of animal names.
+
+export const alphabetizePets = (petArray) => {
+    const petNameArray = petArray.map(x => x.name)
+    const alphabetizedArray = petNameArray.sort(function (a, b) {
+        return a.toLowerCase().localeCompare(b.toLowerCase())
+    })
```
frankmeza: image

You could do something like this:
1. when creating `petNameArray` from `petArray` you can call `.toLowerCase()` there, and then 
1. only have to call `sort()` to alphabetize things

src/index.ts
```diff
@@ -36,25 +39,105 @@ const pets = [
...
+    return alphabetizedArray
+}
+
+// take the array of pets and return a single object where the eys are the animal name and the values are the corresponding animal object with their danger level multiplied by 2.
+
+export const doubleDanger = (petArray) => {
+    const object = petArray.reduce(
+        (obj, item) => Object.assign(obj, { [item.name]: (item.dangerLevel) * 2 }), {}
+    )
```
frankmeza: This looks good! We normally use a slightly different syntax for reduce, like this:

```javascript
const object = petArray.reduce((acc, item) => {
    return { 
        ...acc,
        [item.name]: (item.dangerLevel) * 2,
    }), {})
```

src/index.ts
```diff
@@ -36,25 +39,105 @@ const pets = [
...
+    return animalArray.reduce(
+        (accumalate, element) => {
+            return accumalate.dangerLevel > element.dangerLevel ? element : accumalate
+        }).name
+}
+
+// // 3. Write a function that returns an object containing only type and danger level of the most dangerous animal.
+
+export const findDangerAndType = (animalArray) => {
+    let mostDangerousAnimal = animalArray.reduce(
```
frankmeza: make this `const`

src/index.ts
```diff
@@ -36,25 +39,105 @@ const pets = [
...
+    )
+    return { dangerLevel: mostDangerousAnimal.dangerLevel, type: mostDangerousAnimal.type }
+}
 
 // Who would win?
 // 1. Write a function that takes two numbers (0-6). The numbers represent the index of the pet in the pets array. Write a function that returns the animal which the highest danger level between the two pets.
+// [rido, Dido]
+
+export const animalFight = (num1, num2, animals) => {
+
```
frankmeza: rm this extra line

src/index.ts
```diff
@@ -36,25 +39,105 @@ const pets = [
...
-    locust: 1,
-    // [pet.type]: count
+
+
+export const animalTypeCountFilter = (petArray) => {
+    return petArray.reduce((acc, pet) => {
+        const { type } = pet
+        const accTypeExists = acc[type]
+        return accTypeExists ? { ...acc, [type]: acc[type] + 1 } : { ...acc, [type]: 1 }
+    }, {})
```
frankmeza: 👏  nice

src/testing.ts
```diff
@@ -0,0 +1,55 @@
+import * as indexStuff from "./index" 
+
+export const testTheThing = (actual, expected) => {
+    return actual === expected
+        ? "Good test"
+        : "Bad test! Expected: " + expected + " but received: " + actual
```
frankmeza: convert to string interpolation

src/index.ts
Dcooley1350: This function was designed to do case-insensitive sorting. It was not specified in the prompt so I decided to sort the data regardless of case, and without lowercasing or capitalizing the elements in the array.

src/index.ts
drewcpete: Changed the syntax

src/index.ts
drewcpete: changed let to const

src/index.ts
drewcpete: Extra line was removed

src/testing.ts
drewcpete: Converted bad test output using string interpolation

PR #6, opened 2021-12-16
EventStream interfaces and project skeletons

ianandrewson:
# What
Added skeleton for eventStream function and required types. 
Added skeleton test file for eventStream. 
Pseudocode for scoreArray written. 
Ready to write tests then TDD.

### DIFF
--- a/src/eventStream.test.ts
+++ b/src/eventStream.test.ts
@@ -1,0 +1,5 @@
+describe("Event Stream coding challenge tests", () => {
+	it("passes", () => {
+		expect(true).toBeTruthy()
+	})
+})
--- a/src/eventStream.test.ts
+++ b/src/eventStream.test.ts
@@ -1,5 +1,5 @@
 describe("Event Stream coding challenge tests", () => {
-	it("passes", () => {
-		expect(true).toBeTruthy()
-	})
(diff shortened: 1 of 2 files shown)


src/eventStream.test.ts
```diff
@@ -0,0 +1,5 @@
+describe("Event Stream coding challenge tests", () => {
+	it("passes", () => {
+		expect(true).toBeTruthy()
```
rachoac: use the formatting tool to avoid excessive space

src/eventStream.ts
```diff
@@ -0,0 +1,26 @@
+type EventType = "screenshot" | "new message" | "view"
+
+interface StreamEvent {
+  timestamp: number,
```
rachoac: these should be readonly, to practice immutability

ianandrewson: Ahhh, alright. Will update.

src/eventStream.test.ts
ianandrewson: Hmmmm I just ran the formatting tool but no changes were made to this file.

Are it statements similar to if statements in that there should be a space after it and before the parenthesis?

PR #8, opened 2021-12-16
Step 1 - Create UserAPI class with stub functions, tests

michmitz:
**1. What this does:**

* Creates a class, UserAPI, with defined type interface
* Creates stub CRUD functions with typed inputs/outputs
* Creates stub tests for each UserAPI function

**2. Ticket:**
Discussed with Aron.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -2,13 +2,8 @@
 import { UserAPI } from "./index"
 
 const users = new UserAPI()
-console.log("USERS", users)
 
 describe('Tests will go here!', () => {
-  it('should pass', () => {
-    expect(true).toBeTruthy()
-  })
-
   it('gets a user by id', () => {
     const result = users.getUserById()
 
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,7 +1,41 @@
 //write tests here
+import { UserAPI } from "./index"
+
+const users = new UserAPI()
+console.log("USERS", users)
 
 describe('Tests will go here!', () => {
   it('should pass', () => {
     expect(true).toBeTruthy()
+  })
+
+  it('gets a user by id', () => {
+    const result = users.getUserById()
+
+    expect(result).toStrictEqual(1) 
+  })
+
+  it('gets all users', () => {
+    const result = users.getUsers()
+
+    expect(result).toStrictEqual(1) 
+  }) 
+
+  it('creates a user', () => {
+    const result = users.createUser()
+
+    expect(result).toStrictEqual(1) 
+  })
+
+  it('deletes a user by id', () => {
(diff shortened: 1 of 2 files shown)


src/index.test.ts
```diff
@@ -1,7 +1,102 @@
...
+        age: 35,
+        color: "blue",
+    }) 
+  })
+
+  it.skip('returns no user found if there is no user matching id', () => {
+    const users = new UserAPI()
+      const result = users.getUserById(10)
+
+      expect(result).toStrictEqual(
```
ianandrewson: Maybe it's best to have the returned result be a falsey value if no user is found?

src/index.test.ts
```diff
@@ -1,7 +1,102 @@
...
+            age: 32,
+            color: "red",
+        },
+    ]) 
+  }) 
+
+  it.skip('creates a user', () => {
+    const users = new UserAPI()
+    const newUser = {
+        id: 4,
```
ianandrewson, line 66: If your class doesn't generate an ID, and an ID is instead supplied by the user, what happens if there is already a user with that ID?

src/index.test.ts
```diff
@@ -1,7 +1,102 @@
 //write tests here
+import { UserAPI } from "./index"
```
ianandrewson, line 3: If you're initializing a user API at the beginning of every test, you can avoid repeating code by initializing it in a beforeEach statement.

src/index.test.ts
```diff
@@ -1,7 +1,102 @@
...
+    const result = users.deleteUserById(1)
+
+    expect(result).toStrictEqual({
+        id: 1,
+        name: "Michelle",
+        age: 30,
+        color: "rainbow",
+    }) 
+  })
+
```
ianandrewson, line 102: What happens if the target user of a delete operation can't be found?

src/index.test.ts
```diff
@@ -1,7 +1,102 @@
...
+  it.skip('updates a user by id', () => {
+    const users = new UserAPI()
+    const result = users.updateUser(1, { 
+        name: "Michelle",
+        age: 29,
+        color: "red",
+    })
+
```
ianandrewson, line 115: What happens if the target user of an update operation can't be found?

src/index.ts
```diff
@@ -1 +1,60 @@
...
+
+interface User {
+    readonly id?: number
+    readonly name: string
+    readonly age: number
+    readonly color: string
+}
+
+export class UserAPI {
+    users: ReadonlyArray<User>
```
ianandrewson: Aron is going to say this should be private.

ianandrewson (commented):
Some tests for error handling are in order, but pretty solid all in all!

PR #10, opened 2021-12-17
Step 2 - Implementing Functions

michmitz:
**1. What this does:**
* Implements CRUD functions for user class:
   - Get user by id
   - Get all users
   - Create user
   - Delete user
   - Update user
* Creates CustomError class with customized status code and message
* Tests edge cases for each function, such as when no user can be found by id.

**2. Ticket:**
Discussed with Aron.

**3. How to test:**
-All tests are currently passing.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -52,7 +52,7 @@ describe('Tests will go here!', () => {
       } catch (error: any) {
           expect(error).toBeInstanceOf(CustomError)
           expect(error.message).toBe("No user found.")
-          expect(error.status).toEqual(500)
+          expect(error.status).toEqual(404)
       }
   }) 
 
@@ -95,7 +95,7 @@ describe('Tests will go here!', () => {
     } catch (error: any) {
         expect(error).toBeInstanceOf(CustomError)
         expect(error.message).toBe("No users found.")
-        expect(error.status).toEqual(500)
+        expect(error.status).toEqual(404)
     }
   })
 
@@ -130,7 +130,7 @@ describe('Tests will go here!', () => {
     } catch (error: any) {
         expect(error).toBeInstanceOf(CustomError)
         expect(error.message).toBe("User with id already exists.")
-        expect(error.status).toEqual(500)
+        expect(error.status).toEqual(405)
     } 
   })
 
@@ -153,7 +153,7 @@ describe('Tests will go here!', () => {
       } catch (error: any) {
           expect(error).toBeInstanceOf(CustomError)
           expect(error.message).toBe("No user with that id found.")
-          expect(error.status).toEqual(500)
+          expect(error.status).toEqual(404)
       } 
   })  
 
@@ -187,7 +187,7 @@ describe('Tests will go here!', () => {
       } catch (error: any) {
           expect(error).toBeInstanceOf(CustomError)
           expect(error.message).toBe("Different user with same id already exists.")
-          expect(error.status).toEqual(500)
+          expect(error.status).toEqual(405)
       } 
   })
 
@@ -206,7 +206,8 @@ describe('Tests will go here!', () => {
       } catch (error: any) {
           expect(error).toBeInstanceOf(CustomError)
           expect(error.message).toBe("No user found by that id.")
-          expect(error.status).toEqual(500)
+          expect(error.status).toEqual(404)
       } 
   })
 })
+ 
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -150,8 +150,15 @@ describe('Tests will go here!', () => {
         age: 29,
         color: "red",
     })
-
-    expect(result).toStrictEqual([
+    
+    expect(result).toStrictEqual({
+        id: 1,
+        name: "Michelle",
+        age: 29,
+        color: "red",
+    })
+    
+    expect(users.getUsers()).toStrictEqual([
         {
             id: 1,
             name: "Michelle",
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -157,29 +157,56 @@ describe('Tests will go here!', () => {
       } 
   })  
 
-//   it.skip('updates a user by id', () => {
-//     const result = users.updateUser(1, { 
-//         name: "Michelle",
-//         age: 29,
-//         color: "red",
-//     })
+  it('updates a user by id', () => {
+    const result = users.updateUser(1, { 
+        name: "Michelle",
+        age: 29,
+        color: "red",
+    })
 
-//     expect(result).toStrictEqual({
-//         id: 1,
-//         name: "Michelle",
-//         age: 29,
-//         color: "red",
-//     }) 
-//   })
+    expect(result).toStrictEqual({
+        id: 1,
+        name: "Michelle",
+        age: 29,
+        color: "red",
+    }) 
+  })
 
-//   it.skip('returns null if user with same id already exists', () => {
-//     const result = users.updateUser(1, { 
-//         id: 2,
-//         name: "Michelle",
-//         age: 29,
-//         color: "red",
-//     })
+  it('returns CustomError if user with same id but different name already exists', () => {
+    const updatedUser = {
+        id: 1,
+        name: "John",
+        age: 29,
+        color: "red",
+    }
+    
+    expect(() => users.updateUser(1, updatedUser)).toThrow()
 
-//     expect(result).toStrictEqual(null) 
-//   })
+    try {
+        users.updateUser(1, updatedUser)
+      } catch (error: any) {
+          expect(error).toBeInstanceOf(CustomError)
+          expect(error.message).toBe("Different user with same id already exists.")
+          expect(error.status).toEqual(500)
+      } 
+  })
+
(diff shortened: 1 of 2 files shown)


src/index.test.ts
```diff
@@ -17,15 +43,20 @@ describe('Tests will go here!', () => {
...
-      expect(result).toStrictEqual(null)
-  })
+  it('returns an error if there is no user matching id', () => {
+      expect(() => users.getUserById(10)).toThrow()
+      try {
+           users.getUserById(10)
+      } catch (error: any) {
+          expect(error).toBeInstanceOf(CustomError)
+          expect(error.message).toBe("No user found.")
+          expect(error.status).toEqual(500)
```
ianandrewson: I'm no expert in this, but I'm not sure status 500 is the most semantic status you can use here. On the one hand, it is the server running the code to find the user. On the other hand, it's a specific resource that can't be found, likely queried by the client. Aron should probably weigh in on what the best status codes to use here are. Personally, I implemented error codes in the 400 range.

src/index.test.ts
```diff
@@ -55,13 +86,20 @@ describe('Tests will go here!', () => {
     ]) 
   }) 
 
-  it.skip('returns null if no users are found', () => {
-      const result = users.getUserById(10)
+  it('returns CustomError if no users are found', () => {
```
ianandrewson: The behavior here would probably be defined by a product owner or someone other than you or I, but there is a semantic argument to be made that when getting all users, returning an empty is totally fine instead of returning an error. Technically, all users were found, there just aren't any. And as long as you return an empty array and not null, any front end implementation shouldn't choke on an empty array.

So the question becomes, is finding no users an error?

src/index.test.ts
```diff
@@ -55,13 +86,20 @@ describe('Tests will go here!', () => {
...
+    const noUsers = new UserAPI([])
+    expect(() => noUsers.getUsers()).toThrow()
 
-      expect(result).toStrictEqual(null)
+    try {
+      noUsers.getUsers()
+    } catch (error: any) {
+        expect(error).toBeInstanceOf(CustomError)
+        expect(error.message).toBe("No users found.")
+        expect(error.status).toEqual(500)
```
ianandrewson: A more specific status code can likely be given here as well.

src/index.test.ts
```diff
@@ -78,18 +116,25 @@ describe('Tests will go here!', () => {
...
+    }
 
-      expect(result).toStrictEqual(null)
+    expect(() => users.createUser(newUser)).toThrow()
+      try {
+      users.createUser(newUser)
+    } catch (error: any) {
+        expect(error).toBeInstanceOf(CustomError)
+        expect(error.message).toBe("User with id already exists.")
+        expect(error.status).toEqual(500)
```
ianandrewson: Here too.

src/index.test.ts
```diff
@@ -98,15 +143,21 @@ describe('Tests will go here!', () => {
...
-  it.skip('updates a user by id', () => {
+  it('returns CustomError if deleted user id does not exist', () => {
+    expect(() => users.deleteUserById(10)).toThrow()
+    
+    try {
+        users.deleteUserById(10)
+      } catch (error: any) {
+          expect(error).toBeInstanceOf(CustomError)
+          expect(error.message).toBe("No user with that id found.")
+          expect(error.status).toEqual(500)
```
ianandrewson: And here.

src/index.test.ts
```diff
@@ -121,14 +172,41 @@ describe('Tests will go here!', () => {
...
+    }
+    
+    expect(() => users.updateUser(1, updatedUser)).toThrow()
+
+    try {
+        users.updateUser(1, updatedUser)
+      } catch (error: any) {
+          expect(error).toBeInstanceOf(CustomError)
+          expect(error.message).toBe("Different user with same id already exists.")
+          expect(error.status).toEqual(500)
```
ianandrewson: Yup, and here.

src/index.test.ts
```diff
@@ -121,14 +172,41 @@ describe('Tests will go here!', () => {
...
+    }
+    
+    expect(() => users.updateUser(10, updatedUser)).toThrow()
+
+    try {
+        users.updateUser(10, updatedUser)
+      } catch (error: any) {
+          expect(error).toBeInstanceOf(CustomError)
+          expect(error.message).toBe("No user found by that id.")
+          expect(error.status).toEqual(500)
```
ianandrewson: The last one.

src/index.ts
```diff
@@ -7,54 +7,69 @@ export interface User {
...
+    readonly message: string
+}
+
+export class CustomError extends Error {
+    status: number
+    
+    constructor(status: number, message: string) {
+      super(message);
+      this.status = status
+      Object.setPrototypeOf(this, CustomError.prototype)
```
ianandrewson, line 20: Very curious why you included this. I have no idea if it's right or wrong or anything; this isn't something I did in my implementation and I'm curious as to why you did this.

src/index.ts
```diff
@@ -7,54 +7,69 @@ export interface User {
...
+    }  
 
-    public deleteUserById = (id: number): User | null => {
-        console.log("Create user by id")
-    }
+    public deleteUserById = (id: number): User => {
+        const user = this.users.find(existingUser => existingUser.id === id)
+        
+        if (user) {
+            new UserAPI(this.users.filter(user => user.id !== id))
```
ianandrewson: I still firmly believe there is a way to do this without instantiating a new version of the class.

src/index.ts
```diff
@@ -7,54 +7,69 @@ export interface User {
...
+        let existingUser = this.users.find(existingUser => existingUser.id === id)
 
-    public updateUser = (id: number, user: User): User | null => {
-        console.log("Update user")
+        if (existingUser) {
+            if (existingUser.name === updatedUser.name) {
+                existingUser = { ...updatedUser }
+                return { id, ...existingUser }
+            } else 
+            throw new CustomError(500, "Different user with same id already exists.")
```
ianandrewson: Clever.

ianandrewson (commented):
Super solid. The only thing I would change are the status code if you want to, and not instantiating a new class in your create and delete methods.

src/index.ts
```diff
@@ -7,54 +7,69 @@ export interface User {
...
-        ]
+    constructor(users: ReadonlyArray<User>) {
+        this.users = users
+    }
+
+    public getUserById = (id: number): User => {
+        const foundUser = this.users.find(user => user.id === id)
+        if (foundUser === undefined) {
+            throw new CustomError(500, "No user found.")
+        } else return foundUser
```
rachoac: Braces

src/index.ts
```diff
@@ -7,54 +7,69 @@ export interface User {
...
-                color: "red",
-            },
-        ]
+    constructor(users: ReadonlyArray<User>) {
+        this.users = users
+    }
+
+    public getUserById = (id: number): User => {
+        const foundUser = this.users.find(user => user.id === id)
+        if (foundUser === undefined) {
```
rachoac: !foundUser

src/index.ts
```diff
@@ -7,54 +7,69 @@ export interface User {
...
+    public createUser = (user: User): User => {
+        const userIdExists = this.users.find(existingUser => existingUser.id === user.id)
 
-    public createUser = (user: User): User | null => {
-        console.log("Create user", user)
-    }
+        if (userIdExists) {
+            throw new CustomError(500, "User with id already exists.")
+        } else { 
+            new UserAPI([user, ...this.users])
```
rachoac: Why is this returning a new instance of the api

rachoac (changes requested):

src/index.ts
```diff
@@ -7,54 +7,69 @@ export interface User {
...
+    }  
 
-    public deleteUserById = (id: number): User | null => {
-        console.log("Create user by id")
-    }
+    public deleteUserById = (id: number): User => {
+        const user = this.users.find(existingUser => existingUser.id === id)
+
+        if (user) {
+            this.users = this.users.filter(user => user.id !== id)
```
ianandrewson, line 59: Nice

src/index.ts
```diff
@@ -7,54 +7,78 @@ export interface User {
...
-    }
+    public deleteUserById = (id: number): User => {
+        const user = this.users.find(existingUser => existingUser.id === id)
 
-    public updateUser = (id: number, user: User): User | null => {
-        console.log("Update user")
+        if (user) {
+            this.users = this.users.filter(user => user.id !== id)
+            return user
+        } else throw new CustomError(404, "No user with that id found.")
```
rachoac, line 61: braces

src/index.ts
```diff
@@ -7,54 +7,78 @@ export interface User {
...
+    } 
+
+    public updateUser = (id: number, updatedUser: User): User => {
+        let targetUser = this.users.find(existingUser => existingUser.id === id)
+
+        if (targetUser) {
+            this.users = this.users.map(user => {
+                    if (user.id === id) {
+                        if (user.name === updatedUser.name) {
```
rachoac, line 70: cool

PR #11, opened 2021-12-24
EventScoreAPI

michmitz:
**What this does:**
* Creates a class, EventScoreAPI
* EventScoreAPI has two properties: score and eventArr
* Method - calcHighestScoreArrSequence - returns a contigous sub array derived from eventArr property, no greater than five elements, with the highest score. Score is determined by event type on event element:
  - screenshot = 3
  - views = 2
  - new message = 1

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -372,12 +372,12 @@ describe('EventScore API', () => {
 
         expect(result).toStrictEqual([
             {
-                timestamp: 123123125,
-                eventType: "screenshot",
+                timestamp: 123123124,
+                eventType: "view",
             },
             {
                 timestamp: 123123125,
-                eventType: "new message",
+                eventType: "screenshot",
             },
             {
                 timestamp: 123123125,
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,5 +1,5 @@
-//write tests here
-import { CustomError, UserAPI } from "./index"
+// //write tests here
+import { CustomError, EventScoreAPI, UserAPI } from "./index"
 
 describe('Tests will go here!', () => {
     let users: UserAPI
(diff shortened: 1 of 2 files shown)


src/index.test.ts
```diff
@@ -223,4 +223,116 @@ describe('Tests will go here!', () => {
       } 
   })
 })
- 
\ No newline at end of file
+
+describe('EventScore API', () => {
+    it('Returns an entire array if it is five events or less', () => {
```
ianandrewson, line 228: Might be worth having tests for edge cases: arrays of length 0 and 5.

src/index.ts
```diff
@@ -82,3 +82,56 @@ export class UserAPI {
...
+            this.eventArr.map((event, i) => {
+                const { eventType } = event
+                if (eventType === "screenshot") {
+                    tempCount += 3
+                    } else if (eventType === "view") {
+                        tempCount += 2
+                    } else tempCount += 1
+            
+                if (i > 4) {
+                    if (eventType[i - 1] === "screenshot") {
```
ianandrewson: Are you sure you want to subtract from tempCount the most recent value added?

src/index.ts
```diff
@@ -82,3 +82,56 @@ export class UserAPI {
...
+        let tempCount: number = 0
+        let highestScoreArr: ReadonlyArray<Event> = []
+
+        if (this.eventArr.length <= 5) {
+            highestScoreArr = this.eventArr
+            return highestScoreArr
+        } else {
+            this.eventArr.map((event, i) => {
+                const { eventType } = event
+                if (eventType === "screenshot") {
```
ianandrewson, line 114: Yup, can definitely refactor the conversion of eventType to a scoring value.

src/index.ts
```diff
@@ -82,3 +82,56 @@ export class UserAPI {
...
+    private eventArr: ReadonlyArray<Event>
+
+    constructor(eventArr: ReadonlyArray<Event>) {
+        this.eventArr = eventArr
+        this.score = 0
+    }
+
+    public calcHighestScoreArrSequence = (): ReadonlyArray<Event> => {
+        let tempCount: number = 0
+        let highestScoreArr: ReadonlyArray<Event> = []
```
ianandrewson, line 106: While insignificant in this case, as highestScoreArr is going to have max length of 5, it is possible to solve this without storing the returned highestScoreArr, and instead just directly return it.

ianandrewson (commented):
Definitely on the right track with your algorithm.

PR #12, opened 2022-05-17
Skeleton of an UserAPI for TypeScript-starter-lab

kakizaki55:
made a skeleton of a API for doing basic CRUD operations. 

there are six methods :
-addUser
-getUserById
-updateUserById
-getUsers
-deleteUserById
-searchUserByName

they all console.log the correct descriptions of what each method should do.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,13 +2,6 @@
 console.log("hello there!");
 
 class UserAPI {
-  //constructor for creating each user object for later
-  constructor() {
-    this.id;
-    this.name;
-    this.favoriteColor;
-    this.age;
-  }
   // these are the methods needed for basic CRUD
   addUser(user) {
     console.log(`adds a new user with the name ${user.name}`);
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -21,9 +30,12 @@ class UserAPI {
...
 
+const userObject = { name: "minoka", age: 31, favoriteColor: "green" };
+
 const users = new UserAPI();
-users.addUser("minoka");
+users.addUser(userObject);
 users.getUserById(3);
+users.updateUserById(3);
 users.getUsers();
 users.deleteUserById(1);
```
rachoac, line 32: I don't see implementations for these last two functions in the class above

src/index.ts
```diff
@@ -2,14 +2,23 @@
 console.log("hello there!");
 
 class UserAPI {
-  constructor() {}
+  //constructor for creating each user object for later
+  constructor() {
+    this.id;
```
rachoac: these code has no effect

PR #13, opened 2022-05-17
added addUser and getUserById functionality 

kakizaki55:
**1. What this does:**
Crud Routes for adding and Finding a User 
* addUser will check to see if there is at least a name to create a new user object, will throw an error otherwise. 
* findUserById will check to see if there is a user with that Id, will throw an error otherwise. 
* randomId function is there to generate a unique key for the id,  dynamically. 

**2. Ticket:**
Issue # none

**3. Media:**
Add User | find User By Id |
:-------:|
Screen Shot 2022-05-18 at 4 00 43 PM
Screen Shot 2022-05-18 at 3 57 30 PM
Screen Shot 2022-05-18 at 3 58 04 PM

**4. How to test:**
pull down the code and check to see if the console.log() are identical to screen shots.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,13 +2,6 @@
 console.log("hello there!");
 
 class UserAPI {
-  //constructor for creating each user object for later
-  constructor() {
-    this.id;
-    this.name;
-    this.favoriteColor;
-    this.age;
-  }
   // these are the methods needed for basic CRUD
   addUser(user) {
     console.log(`adds a new user with the name ${user.name}`);
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,7 +18,11 @@ class UserAPI {
   }
   getUserById(id) {
     const targetUser = this.list.filter((item) => item.id === id);
-    return targetUser;
+    if (!targetUser.name) {
+      return "there is no one found with that id";
+    } else {
+      return targetUser;
+    }
   }
   updateUserById(id) {
     console.log(`updates user with Id${id}`);
@@ -38,6 +42,7 @@ const userObject = { name: "minoka", age: 31, favoriteColor: "green" };
 const users = new UserAPI(mockUsers);
 console.log("addUser", users.addUser(userObject));
 console.log("getUserById", users.getUserById(3));
+console.log("getUserById error", users.getUserById(5));
 users.updateUserById(3);
 users.getUsers();
 users.deleteUserById(1);
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,11 +17,11 @@ class UserAPI {
     return newUser;
   }
   getUserById(id) {
-    const targetUser = this.list.filter((item) => item.id === id);
-    if (!targetUser.name) {
+    const targetUser = this.list.filter((user) => user.id === id);
+    if (!targetUser[0]) {
       return "there is no one found with that id";
     } else {
-      return targetUser;
+      return targetUser[0];
     }
   }
   updateUserById(id) {
@@ -40,9 +40,9 @@ class UserAPI {
 const userObject = { name: "minoka", age: 31, favoriteColor: "green" };
 
 const users = new UserAPI(mockUsers);
-console.log("addUser", users.addUser(userObject));
-console.log("getUserById", users.getUserById(3));
-console.log("getUserById error", users.getUserById(5));
+console.log("addUser:", users.addUser(userObject));
+console.log("getUserById:", users.getUserById(3));
+console.log("getUserById error:", users.getUserById(5));
 users.updateUserById(3);
 users.getUsers();
 users.deleteUserById(1);
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,28 +16,20 @@ class UserAPI {
   }
   // these are the methods needed for basic CRUD
   addUser(user) {
-    try {
-      if (!user.name) {
-        throw new Error("you need to at least have a name to add a user")
-      }
-      const newUser = { id: this.randomId(), ...user }
-      this.list.push(newUser)
-      return newUser
-    } catch (err) {
-      console.log(err.message)
+    if (!user.name) {
+      throw new Error("you need to at least have a name to add a user")
     }
+    const newUser = { id: this.randomId(), ...user }
+    this.list.push(newUser)
+    return newUser
   }
 
   getUserById(id) {
     const targetUser = this.list.find((user) => user.id === id)
-    try {
-      if (!targetUser) {
-        throw new Error("There are no users found with that id.")
-      }
-      return targetUser
-    } catch (err) {
-      console.log(err.message)
+    if (!targetUser) {
+      throw new Error("There are no users found with that id.")
     }
+    return targetUser
   }
 
   updateUserById(id) {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,14 +1,31 @@
...
+  { id: 0, name: "minoka", age: 31, favoriteColor: "green" },
+  { id: 1, name: "ted", age: 23, favoriteColor: "black" },
+  { id: 2, name: "ron", age: 41, favoriteColor: "blue" },
+  { id: 3, name: "tim", age: 27, favoriteColor: "red" },
+];
 
 class UserAPI {
-  constructor() {}
+  constructor(users) {
+    this.list = users;
```
rachoac: make passing this list of users optional
the class should maintain its own initially empty structure for holding users, and prepopulate it with the optionally passed in mock users

src/index.ts
```diff
@@ -1,14 +1,31 @@
...
+  }
   // these are the methods needed for basic CRUD
-  addUser(name) {
-    console.log(`inside addUser ${name}`);
+  addUser(user) {
+    const newUser = { id: this.list.length, ...user };
```
rachoac: think of some better way to create a synthetic, random id

src/index.ts
```diff
@@ -1,14 +1,31 @@
...
+    this.list.push(newUser);
+    return newUser;
   }
   getUserById(id) {
-    console.log(`looking for user with Id of ${id}`);
+    const targetUser = this.list.filter((user) => user.id === id);
+    if (!targetUser[0]) {
+      return "there is no one found with that id";
```
rachoac: throw exceptions instead of returning string result. look at the documentation for exceptions

src/index.ts
```diff
@@ -1,14 +1,31 @@
...
-    console.log(`inside addUser ${name}`);
+  addUser(user) {
+    const newUser = { id: this.list.length, ...user };
+    this.list.push(newUser);
+    return newUser;
   }
   getUserById(id) {
-    console.log(`looking for user with Id of ${id}`);
+    const targetUser = this.list.filter((user) => user.id === id);
+    if (!targetUser[0]) {
```
rachoac: is there a different method you can use other than .filter to get just one out?

src/index.ts
```diff
@@ -1,14 +1,35 @@
...
+  addUser(user) {
+    const newUser = { id: this.randomId(), ...user };
+    this.list.push(newUser);
+    return newUser;
   }
   getUserById(id) {
-    console.log(`looking for user with Id of ${id}`);
+    const targetUser = this.list.find((user) => user.id === id);
+    if (!targetUser) {
+      return new Error("There are no users found with that id.");
```
rachoac: read up on throwing exceptions in javascript rather than returning an object. this function should return a user or throw an exception

src/index.ts
```diff
@@ -1,14 +1,35 @@
...
+    } else {
```
rachoac: redundant else since line 28 is the last statement

src/index.ts
```diff
@@ -1,14 +1,35 @@
 //Define class here
-console.log("hello there!");
+const mockUsers = [
```
rachoac, line 2: avoid trailing semi colons, not needed for JSX / typescript

src/index.ts
kakizaki55: as far as throwing an error goes, I cant think of a way to console.log() the thrown error. but when i returned the err in each try catch the correct message was displaying.

src/index.ts
```diff
@@ -1,29 +1,70 @@
...
+  //random id generator
+  randomId = () => {
+    return Math.random().toString(24).slice(2)
+  }
   // these are the methods needed for basic CRUD
-  addUser(name) {
-    console.log(`inside addUser ${name}`);
+  addUser(user) {
+    try {
+      if (!user.name)
```
rachoac: always use braces

src/index.ts
```diff
@@ -1,29 +1,70 @@
...
+        throw new Error("you need to at least have a name to add a user")
+      const newUser = { id: this.randomId(), ...user }
+      this.list.push(newUser)
+      return newUser
+    } catch (err) {
+      err.status = 406
```
rachoac: don't mutate objects that are passed in -- review immutability and javascript

src/index.ts
```diff
@@ -1,29 +1,70 @@
...
+    }
   }
   getUserById(id) {
-    console.log(`looking for user with Id of ${id}`);
+    const targetUser = this.list.find((user) => user.id === id)
+    try {
+      if (!targetUser) throw new Error("There are no users found with that id.")
```
rachoac: always use braces

src/index.ts
```diff
@@ -1,29 +1,70 @@
...
+      const newUser = { id: this.randomId(), ...user }
+      this.list.push(newUser)
+      return newUser
+    } catch (err) {
+      err.status = 406
+    }
   }
   getUserById(id) {
-    console.log(`looking for user with Id of ${id}`);
+    const targetUser = this.list.find((user) => user.id === id)
```
rachoac, line 28: if there is only one argument and its not typed, no need for parens

src/index.ts
```diff
@@ -1,29 +1,70 @@
...
-users.getUserById(3);
-users.getUsers();
-users.deleteUserById(1);
-users.searchUserByName("mino");
+const emptyUsers = new UserAPI([])
+emptyUsers.addUser(userObject)
+emptyUsers.addUser(userObject)
+emptyUsers.addUser(userObject)
+console.log("emptyUsers", emptyUsers)
+// users.updateUserById(3);
```
rachoac, line 76: where's an example of throwing an exception in your 2 functions? use try catch to avoid blowing up the whole program

src/index.ts
```diff
@@ -1,29 +1,70 @@
 //Define class here
-console.log("hello there!");
+const mockUsers = [
+  { id: 0, name: "minoka", age: 31, favoriteColor: "green" },
+  { id: 1, name: "ted", age: 23, favoriteColor: "black" },
+  { id: 2, name: "ron", age: 41, favoriteColor: "blue" },
+  { id: 3, name: "tim", age: 27, favoriteColor: "red" },
+]
```
DuHerb, line 7: For future reference:
folder/file structure wise, we want to try and keep mock data, configuration data/objects, and helper functions in different files.  There are exceptions, of course, but for the most part we try to keep pure data ,view logic, business logic, and reusable functions in different files.  We've established some patterns for this folder structure that you'll see when you start touching real apps.

For this labratory setting though, this is fine.

src/index.ts
DuHerb: let me know if you have questions about using the spread operator to copy/update objects

src/index.ts
```diff
@@ -1,29 +1,70 @@
...
-    console.log(`looking for user with Id of ${id}`);
+    const targetUser = this.list.find((user) => user.id === id)
+    try {
+      if (!targetUser) throw new Error("There are no users found with that id.")
+      return targetUser
+    } catch (err) {
+      err.status = 404
+    }
+  }
+  updateUserById(id) {
```
DuHerb, line 35: For readability, I always add a line in between function/method declarations.  Makes it easier to see where one function ends and the next begins.

src/index.ts
kakizaki55: so I changed the how the API works a bit. I previously had the try catch blocks in the methods them selfs. I think what you wanted was the invoking the the method would crash the app instead. i hope this is right, let me know if there is anything else i can do to clean this up.

PR #15, opened 2022-05-18
Update user by id

kakizaki55:
**1. What this does:**
this function allows you to update an existing user object. 
* if there is no id given it will throw an error.
* if there is no match for the user Id it will throw an error. 
* it will return the newly added user object.

**2. Ticket:**
Issue # none

**3. Media:**
Update by user id |
:-------:|
Screen Shot 2022-05-18 at 4 31 28 PM
Screen Shot 2022-05-18 at 4 31 49 PM
Screen Shot 2022-05-18 at 4 32 04 PM

**4. How to test:**
will have to pull down code and compare console.logs with the screen shots.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,7 +47,7 @@ class UserAPI {
       })
 
       this.list = updatedList
-      const targetUser = this.list.find((user) => user.id === id)
+      const targetUser = this.list.find((user) => user.id === updatedUser.id)
       if (!targetUser) throw new Error("There are no users found with that id.")
       return targetUser
     } catch (err) {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -49,31 +59,54 @@ class UserAPI {
...
+//   console.log(err.message)
+// }
+console.log("-------------------------------------")
+// ---TEST FOR UPDATING USER BY ID METHODS
+console.log(users.updateUserById(updatingUserObject))
 console.log("-------------------------------------")
-console.log("addUser:", users.addUser(userObject))
 try {
-  users.addUser({})
+  users.updateUserById(updatingUserObjectError)
```
rachoac, line 102: we shouldn't be replacing th eexisting add user and update user tests... but instead add new ones for the update

PR #20, opened 2022-05-20
Converting to unit tests

kakizaki55:
**1. What this does:**
* converts all console.log "tests" to actual unit tests
* 14 out of 14 assertions passing. 
* solved the problem for testing when an error is not being thrown by my methods. 
    * this way the assertion will happen before the error is/isnt being thrown. 
 
**2. Ticket:**
Issue # none

**3. Media:**
Test suit status: |
:-------:|
Screen Shot 2022-05-23 at 11 12 40 AM
Screen Shot 2022-05-23 at 12 26 38 PM
^made changes based on slack comments

**4. How to test:**
pull down the code and run ```yarn jest --watch```  in the command line.


src/index.test.ts
```diff
@@ -1,7 +1,133 @@
...
+    users = new UserAPI(mockUsers)
+  })
+
+  afterEach(() => {
+    users = null
+  })
+
+  it("should make an instance of the UserAPI class and add a user, new user object should have an id.", () => {
+    const newUser = users.addUser(userObject)
+
```
kakizaki55, line 17: this is the line that not only adds a user but mutates the original mockUsers array.

src/index.test.ts
```diff
@@ -1,7 +1,133 @@
...
+  updatingUserObject,
+  updatingUserObjectError,
+  updatingUserObjectErrorWithId,
+} from "./mockdata"
 
-describe('Tests will go here!', () => {
-  it('should pass', () => {
-    expect(true).toBeTruthy()
+describe("Tests will go here!", () => {
+  let users
```
rachoac: don't have a global users attribute -- each test should be self contained and setup its own data , otherwise tests will step on each other.

if you need to create a utitility that creates test data, make that stateless and invoke it in each individual test to create data for you

kakizaki55: yeah okay, that makes sense.

src/index.test.ts
```diff
@@ -1,7 +1,132 @@
...
+
+    expect(targetUser2).toEqual({
+      id: "3",
+      name: "tim",
+      age: 27,
+      favoriteColor: "red",
+    })
+
+    try {
+      users.addUser({})
```
rachoac: what if no error is thrown? shouldn't we fail?

src/index.test.ts
```diff
@@ -1,7 +1,132 @@
...
+    }
+
+    try {
+      users.addUser()
+    } catch (err) {
+      expect(err.message).toEqual("Cannot read property 'name' of undefined")
+    }
+
+    expect(() => users.addUser()).toThrow(new Error(expect.any(String)))
+  })
```
kakizaki55, line 43: this is the closest solution I could come up with. but it still doesn't work.

src/index.test.ts
```diff
@@ -1,7 +1,140 @@
...
+
+    expect(newUser).toEqual({ id: expect.any(String), ...userObject })
+  })
+
+  it("should show the correct error message, will fail if no errors are thrown", () => {
+    const users = new UserAPI(mockUsers())
+
+    try {
+      users.addUser({})
+      throw new Error("it should not reach this line of code")
```
rachoac: almost right -- we want to find out how to fail a test
not break the test

PR #21, opened 2022-05-20
Convert UserAPI  to Typescript

kakizaki55:
**1. What this does:**
* Added interface for the Person to use as the user Type. 
* Each method has an argument type and return type. 
* Each Array is of type ReadonlyArray 
* had to make changes to line 23, to reassign the array so the Array stays unmutated. 
* added UserFormMeta interface to add user and update user arguments. (id is not optional)
* updateUser now takes 2 arguments (id, UserFormMeta), Updated Tests accordingly. 

**2. Ticket:**
Issue #none

**3. Media:**
 UserFormMeta interface |
:-------:|
Screen Shot 2022-05-24 at 10 01 20 AM

unit tests |
:-------:|
Screen Shot 2022-05-23 at 2 59 32 PM

**4. How to test:**
pull down code run ``` yarn jest --watch```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -79,4 +79,23 @@ export default class UserAPI {
 
     return filteredUserArray
   }
+  getAverageAge() {
+    const ageArray = Object.values(this.list).filter(users => users.age)
+
+    const ageOnlyArray = ageArray.map((users: Person) => {
+      if (users.age) {
+        const age = users.age
+        return age
+      } else {
+        //some how this only work if i return 0 the type for .map is number | undefined.
+        return 0
+      }
+    })
+
+    const averageAge =
+      ageOnlyArray.reduce((prev, current) => prev + current) /
+      ageOnlyArray.length
+
+    return averageAge
+  }
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ export default class UserAPI {
     return Math.random().toString(24).slice(2)
   }
 
-  addUser(user: Person): Person {
+  addUser(user: Person): Person | Error {
     if (!user.name) {
       throw new Error("you need to at least have a name to add a user")
     }
@@ -24,7 +24,7 @@ export default class UserAPI {
     return newUser
   }
 
-  getUserById(id: string): Person {
+  getUserById(id: string): Person | Error {
     const targetUser = this.list.find(user => user.id === id)
     if (!targetUser) {
       throw new Error("There are no users found with that id.")
@@ -32,7 +32,7 @@ export default class UserAPI {
     return targetUser
   }
 
-  updateUserById(updatedUser: Person): Person {
+  updateUserById(updatedUser: Person): Person | Error {
     if (!updatedUser.id) {
       throw new Error("We can not update a user without an id")
     }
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -1,15 +1,21 @@
...
   }
-  //random id generator
-  randomId = () => {
+
+  randomId = (): string => {
     return Math.random().toString(24).slice(2)
   }
 
-  addUser(user) {
+  addUser(user: Person): Person | Error {
```
rachoac: error is never returned, so its not appropriate to declare this

src/index.ts
```diff
@@ -1,15 +1,21 @@
...
+interface Person {
+  readonly id?: string
+  readonly name: string
+  readonly age?: number
+  readonly favoriteColor?: string
+}
 
 export default class UserAPI {
-  constructor(users) {
+  list: Array<Person>
```
rachoac: Array -> ReadonlyArray

src/index.ts
```diff
@@ -40,18 +46,18 @@ export default class UserAPI {
     return this.getUserById(updatedUser.id)
   }
 
-  getUsers() {
+  getUsers(): Array<Person> | [] {
```
rachoac: Array -> ReadonlyArray

src/index.ts
```diff
@@ -1,32 +1,38 @@
-//@ts-nocheck
+interface Person {
+  readonly id?: string
```
DuHerb: is there a reason `id` is optional?  

In general, if an object exists, it should have an `id`, especially if you're using an array to store the object in. the id is your unique identifier for indexing/lookups.

DuHerb (changes requested):

src/index.ts
```diff
@@ -1,32 +1,38 @@
...
   }
-  //random id generator
-  randomId = () => {
+
+  randomId = (): string => {
     return Math.random().toString(24).slice(2)
   }
 
-  addUser(user) {
+  addUser(user: Person): Person {
```
DuHerb: instead of `Person`, create a `userFormData` or equivalent shape that represents required/optional initialization fields.  Then your function 'handles' that data, (in production code, it would be good to have a pure function handle this), and provides the `Person` object to whatever api is handling storage.

src/index.ts
```diff
@@ -1,32 +1,38 @@
...
+  getUserById(id: string): Person {
     const targetUser = this.list.find(user => user.id === id)
     if (!targetUser) {
       throw new Error("There are no users found with that id.")
     }
     return targetUser
   }
 
-  updateUserById(updatedUser) {
+  updateUserById(updatedUser: Person): Person {
```
DuHerb: updatedUserById should accept:
a) a user ID for lookup
b) fields to be updated

lookup the TS Utility `Partial`, or just create a new object with fields that may be updated.

DuHerb: @rachoac should this be refactored to actually update the found `Person` object here, rather than accepting a whole, new `Person`?  @kakizaki55 and I have discussed 1-1 in zoom, so maybe not necessary to practice that pattern here.

PR #23, opened 2022-05-24
Convert to Record Object. 

kakizaki55:
**1. What this does:**
* converts argument data to Typed Record Object. 
* All test still passing. 
* Small adjustment to test to work with new data model 
* refactored and added an interface utils file. 
* refactored getUserById to not use a .find
* refactored updateUser with spread op 
* refactored getUsers with .map 
* refactored deleteUser with spread op
* refactored searchUser with .filter
* all iterations are typed. 

**2. Ticket:**
Issue # none

**3. Media:**
Screen Shot 2022-05-24 at 4 14 17 PM
Screen Shot 2022-05-24 at 4 14 33 PM
Screen Shot 2022-05-24 at 3 58 49 PM

**4. How to test:**
pull down code and run ```yarn jest --watch```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -79,4 +79,23 @@ export default class UserAPI {
 
     return filteredUserArray
   }
+  getAverageAge() {
+    const ageArray = Object.values(this.list).filter(users => users.age)
+
+    const ageOnlyArray = ageArray.map((users: Person) => {
+      if (users.age) {
+        const age = users.age
+        return age
+      } else {
+        //some how this only work if i return 0 the type for .map is number | undefined.
+        return 0
+      }
+    })
+
+    const averageAge =
+      ageOnlyArray.reduce((prev, current) => prev + current) /
+      ageOnlyArray.length
+
+    return averageAge
+  }
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -106,26 +106,29 @@ export default class UserAPI {
       .map((users: Person) => {
         onlyColors.add(users.favoriteColor)
       })
+    console.log("onlyColors", onlyColors)
 
     return onlyColors
   }
 
-  getFavoriteColorCount(): object {
+  getFavoriteColorCount(): Record<string, number> {
     const colorCount = Object.values(this.list).reduce(
-      (acc: object, person: Person) => {
-        if (`${person.favoriteColor}` in acc) {
-          //@ts-ignore
-          ++acc[person.favoriteColor]
+      (acc, person: Person) => {
+        if (acc[person?.favoriteColor as string]) {
+          const newColorCount = acc[person.favoriteColor as string] + 1
 
-          return acc
+          return {
+            ...acc,
+            [person.favoriteColor as string]: newColorCount,
+          }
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -1,20 +1,10 @@
...
-  readonly age?: number
-  readonly favoriteColor?: string
-}
+import { Person, UserFormMeta } from "./interfaces"
 
 export default class UserAPI {
-  list: ReadonlyArray<Person>
-  constructor(users: ReadonlyArray<Person>) {
-    this.list = users || []
+  list: Record<string, Person>
```
DuHerb: this can be the `Users` type that you're exporting from your config/mockData file

src/index.ts
```diff
@@ -1,20 +1,10 @@
...
+
+  constructor(users: Record<string, Person>) {
```
DuHerb: same

src/index.ts
```diff
@@ -25,50 +15,70 @@ export default class UserAPI {
     if (!user.name) {
       throw new Error("you need to at least have a name to add a user")
     }
-    const newUser = { id: this.randomId(), ...user }
-    this.list = [...this.list, newUser]
+
+    const id = this.randomId()
+    const newUser = { id, ...user }
+    this.list = { ...this.list, [id]: newUser }
```
DuHerb, line 21: another way we could do this is to have a `private` function to handle actually adding a user to the list.

`addUser` creates the new object and passes it to the private `updateUserList` functions that performs this action.

then we could also use this function for things like `updateUserById`, which would parse/create the new user object, and then pass it to `updateUserList`.

DuHerb: not required to refactor your work here, just something to keep in mind when we start building other API's

src/index.ts
```diff
@@ -25,50 +15,70 @@ export default class UserAPI {
...
+
+    return userArray
   }
 
   deleteUserById(id: string): Person {
     const deletedUser = this.getUserById(id)
-    const newUsersList = this.list.filter(user => user.id !== id)
-    this.list = newUsersList
+
+    delete this.list[id]
```
DuHerb: `delete` is a valid way of removing an item from an object.  
But in keeping with the theme of `immutability`, how might this be done using the `...` syntax?

remember to check that the property you are targeting exists on the object first.

src/index.ts
```diff
@@ -25,50 +15,70 @@ export default class UserAPI {
...
     return deletedUser
   }
 
   searchUserByName(name: string): ReadonlyArray<Person> | [] {
-    const filteredUserArray = this.list.filter(user => {
-      if (user.name.toLowerCase().includes(name.toLowerCase())) {
-        return user
+    const filteredUserArray = []
+
+    for (const [key, value] of Object.entries(this.list)) {
```
DuHerb: a cleaner way of doing this search would be to use the `Array.filter` method. 

You can `chain` methods together here.  Investigate other `Object` methods on mdn that can be used here.

DuHerb (approved):

PR #27, opened 2022-06-07
create userapi skeleton

Plymatea:
**1: Ticket** 
None

**2: What it does**
Creates a class "UserAPI" with the skeleton structure  with the following methods
- Constructs an empty array
- assignID() (for future use when adding users)
- addUser(user)
- getUserById(id)
- getUsers()
- deleteUsersById(id)
- searchUserByName(name)

I was unsure if the User was going to be a separate class or baked into the UserAPI class. Decided to create a second User class and can refactor if needed
The provided index file was typescript format, so I added in the data verification for the variables instead of changing the filetype to  index.js.

**3: Media**
none

**4: Test**
Compiles correctly, Nothing to test with only skeleton

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,9 +4,9 @@ class User {
   age: number
     
   constructor(name: string, color: string, age: number) {
-    this.name = name;
-    this.favColor = color;
-    this.age = age;
+    this.name = name
+    this.favColor = color
+    this.age = age
   }
 }
 
@@ -14,7 +14,7 @@ class UserAPI {
   users: Array<User>
 
   constructor() {
-    this.users = [];
+    this.users = []
   } 
 
   private assignId() {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,1 +1,45 @@
-//Define class here
+class User {
+  name: string;
+  favColor: string;
+  age: number;
+
+    constructor(name: string, color: string, age: number) {
+    this.name = name;
+    this.favColor = color;
+    this.age = age;
+  }
+}
+
+class UserAPI {
+  users: Array<User>;
+
+  constructor() {
+    this.users = []
+  } 
+
+  assignId() {
+
+  }
+
+  addUser(user: User) {
+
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1 +1,45 @@
-//Define class here
\ No newline at end of file
+class User {
+  name: string;
+  favColor: string;
+  age: number;
+
+    constructor(name: string, color: string, age: number) {
```
rachoac: remove leading whitespace

src/index.ts
```diff
@@ -1 +1,45 @@
-//Define class here
\ No newline at end of file
+class User {
+  name: string;
+  favColor: string;
+  age: number;
```
rachoac: no need for semi colons (es6 / typescript conventions)

src/index.ts
```diff
@@ -1 +1,45 @@
...
+}
+
+class UserAPI {
+  users: Array<User>;
+
+  constructor() {
+    this.users = []
+  } 
+
+  assignId() {
```
rachoac: could be private, since its not part of the public api (eg. it shouldn't be accessible by callers outside of the functions of this class)

Plymatea: Cool! I didn't know you could do that!

src/index.ts

src/index.ts
Plymatea: Do you have a link to the es6 / typescript conventions? The internet seems to have competing thoughts on semicolon conventions. Is there a website that lays this out you'd like me to use?

src/index.ts

src/index.ts
```diff
@@ -1 +1,45 @@
...
+    this.favColor = color;
+    this.age = age;
+  }
+}
+
+class UserAPI {
+  users: Array<User>
+
+  constructor() {
+    this.users = [];
```
rachoac: semi colons are still evident in this file, please remove them all

PR #28, opened 2022-06-07
addUser & getUserById methods functionality

Plymatea:
**1: Ticket**
None

**2: What this does:**

- Converts Class User to Interface IUser
- Adds UserAPI property "nextId"
- Add UserAPI private createUser()
- Privatizes UserAPI properties
- Adds functionality to UserAPI addUser() && getUserById()
- Throws error when user submits data with pre-existing ID field
- Throws error when user submits duplicate user to be entered into database. *(checks for equality across name, favcolor, age properties)*

**3: Media**

Code to: 
- create API
- pass user into addUser()
- find user by getUserById()
image- Terminal window showing console.log of code above
image

**4: Test**
Compiles correctly with only expected errors present.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,7 +48,7 @@ class UserAPI {
   
   private createUser(name: string, favColor: string, age: number): IUser {
     const id = this.assignId()
-    return { id, name, favColor, age };
+    return { id, name, favColor, age }
 }
 }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,9 +23,9 @@ class UserAPI {
     const result = (this.users).find( ( { name, favColor, age }: IUser) => (name === user.name && favColor === user.favColor && age === user.age))
     try {
       if (user.id) {
-        throw "Error: User Id is set by database. Please resubmit without pre-existing value."
+        throw new SyntaxError("Please resubmit without pre-exisitng ID field.")
       } else if (result) {
-        throw "Error: A user with those properties already exists in the database."
+        throw new Error("A user with those properties already exists in the database.")
       } 
     } catch (error) {
       return error
@@ -43,10 +43,10 @@ class UserAPI {
       if (foundUser) {
         return foundUser
       }
-      throw "Error: No user found."
+      throw new ReferenceError("No user found.")
     }
     catch (error) {
-      return (error)
+      return error
     }
   }
 
@@ -74,6 +74,11 @@ const user = {
   favColor: "blue",
   age: 247
 }
+const userDuplicate = {
+  name: "andy",
+  favColor: "blue",
+  age: 247
+}
 const user2 = {
   id: 5,
   name: "andy",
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,32 +1,37 @@
...
   private assignId() {
-
+    this.nextId++
+    return this.nextId
   }
 
-  addUser(user: User) {
-
+  addUser(user: IUser) {
+    const {name, favColor, age} = user
```
rachoac: if the user has an id already, this function should throw an exception (see javascript.info for how to throw errors)

src/index.ts
```diff
@@ -1,32 +1,37 @@
...
+    return newUser
   }
 
   getUserById(id: number) {
-
+    const foundUser = this.users.find( user => user.id === id) 
+    if (foundUser) {
+      return foundUser
+    }
+    return "No user found"
```
rachoac: throw an error instead of returning a string or an object

src/index.ts
Plymatea: Can I get some clarifications here: 
1) Will the user object being passed to the API have a pre-existing ID property that should be stored in the database?
1a) If so, the error checking is to verify no two users have the same ID, correct? 
2) You want error checking for duplicate users. ex name: {"andy" age: 111}  => Error: User already exists in the database

rachoac: `Will the user object being passed to the API have a pre-existing ID property that should be stored in the database?` someone calling the addUser method could formulate a user object with the id already set. This is an error condition, becuase we want the userapi to be the one assigning ids.

`You want error checking for duplicate users. ex name: {"andy" age: 111} => Error: User already exists in the database` correct

Plymatea: Thank you

src/index.ts
```diff
@@ -1,32 +1,53 @@
...
-
+    try {
+      const foundUser = this.users.find( user => user.id === id) 
+      if (foundUser) {
+        return foundUser
+      }
+      throw "Error: No user found."
+    }
+    catch (error) {
+      return (error)
```
rachoac: we should not be returning errors ... let the function throw the error and have its caller deal with it

src/index.ts
```diff
@@ -1,32 +1,53 @@
...
-
+  addUser(user: IUser) {
+    const result = (this.users).find( ( { name, favColor, age }: IUser) => (name === user.name && favColor === user.favColor && age === user.age))
+    try {
+      if (user.id) {
+        throw "Error: User Id is set by database. Please resubmit without pre-existing value."
+      } else if (result) {
+        throw "Error: A user with those properties already exists in the database."
+      } 
+    } catch (error) {
```
rachoac: see comment below

src/index.ts
rachoac: be sure to have a case in the console logs below that intentionalyl induces the error and catches for it there

PR #29, opened 2022-06-07
Add functionality for addUser() and getUserById() methods

dlindsey-olioapps:
**Ticket**
Not applicable
	
**What does this PR do**
This PR will add functionality to the addUser() and getUserById() methods. In order to make this work, the private method assignId() has been built out. I've also added in an interface for user instantiation, instead of a class. Code for testing the methods has also been added to the bottom of index.ts.

**Media**
image

**How to test**
To test, clone the repository to your computer and navigate to the project with the terminal. When ready, run _**yarn run dev**_ to execute the application. You should see three tests, the first two with results of a User, and the third test returning "User not found" - similar to the screenshot above.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,33 +2,42 @@ class User {
   name: string
   favoriteColor: string
   age: number
-  id: number
+  id?: number
 
-  constructor(name: string, favoriteColor: string, age: number, id: number) {
+  constructor(name: string, favoriteColor: string, age: number) {
     this.name = name
     this.favoriteColor = favoriteColor
     this.age = age
-    this.id = id
   }
 }
 
+
 class UserAPI {
   users: Array<User>
+  currentId: number
 
   constructor() {
     this.users = []
+    this.currentId = 0
   }
 
   private assignId() {
-
+    this.currentId++;
+    return this.currentId - 1;
   }
 
   addUser(user: User) {
-
+    user.id = this.assignId();
+    this.users.push(user);
   }
 
   getUserById(id: number) {
-
+    const user = this.users.find(x => x.id === id);
+    if (user !== undefined) {
+      return user;
+    } else {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,34 +1,36 @@
...
+    return this.currentId - 1
   }
 
   addUser(user: User) {
-
+    user.id = this.assignId()
+    this.users.push(user)
   }
 
   getUserById(id: number) {
```
rachoac, line 39: rather than return a string or a user object, throw an error if the user is not found (see javascript.info on how to throw errors)

src/index.ts
```diff
@@ -1,34 +1,36 @@
...
 
   private assignId() {
-
+    this.currentId++
+    return this.currentId - 1
   }
 
   addUser(user: User) {
-
+    user.id = this.assignId()
```
rachoac: what if the user is being added with an id that already exists?
in fact, if the user has an id, add should not work (its already added) ... in this case it should throw an error

also ... the added user should be returned

src/index.ts
```diff
@@ -1,34 +1,58 @@
...
+          throw "ERROR: User with these properties already exists"
+        }
+        user.id = this.assignId()
+        this.users.push(user)
+        return user
+      } else {
+        throw "ERROR: Id incorrectly provided by input user"
+      }
+    }
+    catch (error) {
```
rachoac: rather than return a string or a user object, throw an error if the user is not found (see javascript.info on how to throw errors)

src/index.ts
rachoac: make sure to intentionally induce the error in the console log tests below

src/index.ts
rachoac: make sure to intentionally induce the error in the console log tests below

PR #31, opened 2022-06-08
Functionality for getUsers, deleteUserById, searchUserByName, searchUserByFavoriteColor

Plymatea:
**Ticket**
None

**This Does:**
- adds functionality for `getUsers()`
- adds functionality for `deleteUserById()`
- adds functionality for `searchUserByName()`
- adds functionality for `searchUserByFavoriteColor()`
- adds `_` to private properties per convention
- removed `assignId()` method and assoc. properties and integrated with `addUser()`
- removed `createUser()` helper method and integrated with `addUser()`
- converted id property to string from number
-  added one user in dataset with pre-determined id as seed for testing purposes
- `UserAPI()` will accept an optional array in the constructor intended for seed data or users. Otherwise defaults to empty array

**Media**

image

**Test**

Created new console tests new methods. Included `.toLower()` to help with user input variety. 
image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -49,17 +49,18 @@ class UserAPI {
   }
 
   get_Users() {
-    if (this._users.length > 0) {
-      return this._users
+    if (this._users === null || this._users === undefined) {
+      throw new Error(`User Dataset not found`)
     }
-    throw new Error(`No users to return`)
+    return this._users
   }
 
   deleteUserById(id: number) {
-    const indexOfUser = this._users.findIndex( user => user.id === id) 
+    const indexOfUser = this._users.findIndex( user => user.id === id)
+    const user = this.getUserById(id) 
     if (indexOfUser > -1) {
       this._users.splice(indexOfUser,1)
-      return `Deleted user id ${id}`
+      return user
     } else{
       throw new ReferenceError(`No user found with id ${id}.`)
     }  
@@ -68,7 +69,7 @@ class UserAPI {
   searchUserByName(name: string) {
     const users = this._users.filter( user => user.name.toLowerCase() === name.toLowerCase())
     
-    if (users.length > 0) {
+    if (users.length) {
       return users
     } 
     throw new ReferenceError(`No user found with name ${name}`)
@@ -76,7 +77,7 @@ class UserAPI {
 
   search_UsersByFavoriteColor(color: string) {
     const users = this._users.filter( user => user.favColor.toLowerCase() === color.toLowerCase())
-    if (users.length > 0) {
+    if (users.length) {
       return users
     } 
     throw new ReferenceError(`No users favorite color is ${color}`)
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,17 +6,22 @@ interface IUser {
 }
 
 class UserAPI {
-  private users: Array<IUser>
-  private nextId: number
+  private _users: Array<IUser>
+  private _nextId: number
 
   constructor() {
-    this.users = []
-    this.nextId = 0
+    this._users = []
+    this._nextId = 0
   } 
 
   private assignId() {
-    this.nextId++
-    return this.nextId
+    this._nextId++
+    return this._nextId
+  }
+  
+  private createUser(name: string, favColor: string, age: number): IUser {
+    const id = this.assignId()
+    return { id, name, favColor, age }
   }
 
   addUser(user: IUser) {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -6,101 +6,202 @@ interface IUser {
...
-    this.users = []
-    this.nextId = 0
+    this._users = []
+    this._nextId = 0
   } 
 
   private assignId() {
-    this.nextId++
-    return this.nextId
+    this._nextId++
```
DuHerb: obviously this is a simple use case, but a more reliable way of creating unique ideas in development would be something like `Date.now().toString()` or some other similar function that uses `Math.random` or date objects to return strings.

src/index.ts
```diff
@@ -6,101 +6,202 @@ interface IUser {
...
+    throw new ReferenceError(`No user found with id ${id}.`)
   }
 
-  getUsers() {
-
+  get_Users() {
+    if (this._users.length > 0) {
+      return this._users
+    }
+    throw new Error(`No users to return`)
```
DuHerb: empty lists are valid states.  I would say return the empty list here, and let your UI handle it accordingly.

only error here would be if there was no list to return. `undefined` or `null`.  or if you caught an error from an upstream service request (eg: this API makes a `get` request that fails).

src/index.ts
```diff
@@ -6,101 +6,202 @@ interface IUser {
...
   }
 
   deleteUserById(id: number) {
-
+    const indexOfUser = this._users.findIndex( user => user.id === id) 
+    if (indexOfUser > -1) {
+      this._users.splice(indexOfUser,1)
+      return `Deleted user id ${id}`
```
DuHerb: returning the deleted user would be the realworld usecase.  It would contain the `id` as well as any other data the UI might want

src/index.ts
```diff
@@ -6,101 +6,202 @@ interface IUser {
...
+    } else{
+      throw new ReferenceError(`No user found with id ${id}.`)
+    }  
   }
 
   searchUserByName(name: string) {
+    const users = this._users.filter( user => user.name.toLowerCase() === name.toLowerCase())
+    
+    if (users.length > 0) {
```
DuHerb: try `if (user.length)`.  i believe it should work: if user.length is falsy, then we don't have any users to return

src/index.ts
```diff
@@ -6,101 +6,202 @@ interface IUser {
...
+      return users
+    } 
+    throw new ReferenceError(`No user found with name ${name}`)
+  }
 
+  search_UsersByFavoriteColor(color: string) {
+    const users = this._users.filter( user => user.favColor.toLowerCase() === color.toLowerCase())
+    if (users.length > 0) {
```
DuHerb: same.  make sure it passes test.  If not....then whoops! on me and leave as is.

src/index.ts
Plymatea: Changed to verify if list if null or undefined. Those states will throw error, all others will return the list.

src/index.ts
Plymatea: Changed from returning a string to returning the deleted user

src/index.ts

src/index.ts
Plymatea: Done, It seems to work fine with the current tests set up.

src/index.ts
Plymatea: There are a lot of line changes to convert `assignId` to `Date.now()`.  I also removed the `assignId` functionality inside the `addUser` Method.

src/index.ts
```diff
@@ -1,106 +1,189 @@
...
 
 class UserAPI {
-  private users: Array<IUser>
-  private nextId: number
+  private _users: Array<IUser>
 
   constructor() {
-    this.users = []
-    this.nextId = 0
+    this._users = [{ id: "1", name: "andy", favColor: "blue", age: 247 }]
```
DuHerb: i think there was a misunderstanding.  

`UserApi` should accept an optional argument of `Array<IUser>`.  If provided, `this._users` should be initialized with the argument as it's value.  Otherwise, `this._users` should be empty.

when testing, you pass your 'testUsers` when creating a `UserApi`

Plymatea: Yes I misunderstood.  
I've change the constructor to default to an empty [] unless passed an array with seed data.

DuHerb (approved):

PR #32, opened 2022-06-08
Add methods/tests for deleting users and finding by favorite color, name, or all

dlindsey-olioapps:
**Ticket**
Not applicable

**What does this PR do**
This PR will add tests for and functionality to the following methods
- getUsers()
- deleteUserById()
- searchUserByName()
- searchUsersByFavoriteColor()

The tests for each method check for both expected returns and unexpected returns. 

I also have removed the currentId property and am now assigning Ids based on Date.now(). For testing this, I create a user with addUser() and then run the test based on the returned user.id property.

The code has also been updated to conform to standard conventions with regards to naming private class properties. For example, the private number property _users_ has now been renamed from _users_ to __users_

**Media**
image
image
image
image
image

**How to test**
To test, clone the repository to your computer and navigate to the project with the terminal. Download all necessary packages with _yarn install_ and when ready, run **_yarn run dev_** to execute the application. You should see several tests and the results of these tests.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -180,20 +180,15 @@ try {
 }
 console.log("--------------------------------------------------------------------")
 
-const addedUserForIdTest = x.addUser({
-  name: "Jenny",
-  favoriteColor: "yellow",
-  age: 99
-})
 console.log(`Test:
 Should successfully find user based on id
 `)
 console.log(`Expect:
-{name: 'Jenny', favoriteColor: 'yellow', age: 99, id: "${addedUserForIdTest.id}"}
+{name: 'Jenny', favoriteColor: 'yellow', age: 99, id: "1"}
 `)
 console.log("Result")
 try {
-  console.log(x.getUserById(addedUserForIdTest.id as string))
+  console.log(x.getUserById("1"))
 } catch (e: any) {
   console.log(e.message)
 }
@@ -244,20 +239,16 @@ try {
 }
 console.log("--------------------------------------------------------------------")
 
-const addedUserForDeleteTest = x.addUser({
-  name: "James",
-  favoriteColor: "white",
-  age: 99
-})
+
 console.log(`Test:
 Removes user from the UserAPI array
 `)
 console.log(`Expect:
-DELETE: User with id:${addedUserForDeleteTest.id} is removed from following
+DELETE: User with id:2 is removed from following
 `, x.getUsers())
 console.log("Result:")
 try {
-  console.log("Deleted:", x.deleteUserById(addedUserForDeleteTest.id as string))
+  console.log("Deleted:", x.deleteUserById("2"))
   console.log("Remaining:", x.getUsers())
 } catch (e: any) {
   console.log(e.message)
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,22 +6,22 @@ interface User {
 }
 
 class UserAPI {
-  private users: Array<User>
-  private currentId: number
+  private _users: Array<User>
+  private _currentId: number
 
   constructor() {
-    this.users = []
-    this.currentId = 1
+    this._users = []
+    this._currentId = 1
   }
 
   private assignId() {
-    this.currentId++
-    return this.currentId - 1
+    this._currentId++
+    return this._currentId - 1
   }
 
   addUser(user: User) {
     if (!user.hasOwnProperty("id")) {
-      const existingUser = this.users.find(
+      const existingUser = this._users.find(
         x => x.name === user.name
           && x.favoriteColor === user.favoriteColor
           && x.age === user.age)
@@ -29,7 +29,7 @@ class UserAPI {
         throw new Error("User with these properties already exists")
       }
       user.id = this.assignId()
-      this.users.push(user)
+      this._users.push(user)
       return user
     } else {
       throw new Error("Id incorrectly provided by input user")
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -46,15 +46,38 @@ class UserAPI {
   }
 
   getUsers() {
-
+    if (this._users.length > 0) {
+      return this._users;
+    } else {
+      throw new Error("There are no users present")
```
DuHerb: an empty list is a valid `userList` state.
in this case, it should return the empty `userList`, and allow the caller to handle the scenario.

here, I think the only thing that would throw an error is if `users` was somehow not defined.

src/index.ts
```diff
@@ -46,15 +46,38 @@ class UserAPI {
...
+    }
   }
 
   deleteUserById(id: number) {
-
+    const userIndex = this._users.findIndex(x => x.id === id)
+    if (userIndex > 0) {
+      this._users.splice(userIndex, 1)
```
DuHerb, line 54: please return the deleted user

src/index.ts
```diff
@@ -46,15 +46,38 @@ class UserAPI {
...
+    } else {
+      throw new Error("User was not found")
+    }
   }
 
   searchUserByName(name: string) {
+    const userResult = this._users.filter(x => x.name.toLowerCase() === name.toLowerCase())
+    if (userResult.length > 0) {
```
DuHerb: `if (userResult.length)` should be sufficient.  Test to make sure.

src/index.ts
```diff
@@ -46,15 +46,38 @@ class UserAPI {
...
+      return userResult
+    } else {
+      throw new Error("User(s) not found")
+    }
+  }
 
+  searchUsersByFavoriteColor(color: string) {
+    const userResult = this._users.filter(x => x.favoriteColor.toLowerCase() === color.toLowerCase())
+    if (userResult.length > 0) {
```
DuHerb: same

src/index.ts
```diff
@@ -6,38 +6,38 @@ interface User {
...
-    this.users = []
-    this.currentId = 1
+    this._users = []
+    this._currentId = 1
   }
 
   private assignId() {
-    this.currentId++
-    return this.currentId - 1
+    this._currentId++
```
DuHerb: obviously this is a simple use case, but a more reliable way of creating unique ideas in development would be something like `Date.now().toString()` or some other similar function that uses `Math.random` or date objects to return strings.

dlindsey-olioapps: Thanks! I've changed the id to be instead of an incremented number, Date.now (as the id is set as a number type). 

For tests, I create a user and then run the test based on the Id property in the user object that is returned from addUser()

src/index.ts
dlindsey-olioapps: fixed! thanks

src/index.ts
dlindsey-olioapps: fixed! thanks

src/index.ts
dlindsey-olioapps: fixed! thanks

src/index.ts
dlindsey-olioapps: I now have array error handling added, and return an empty [] instead of an error when the array is empty. I was unable to figure out how to test this though without the TSC compiler throwing errors.

src/index.ts
```diff
@@ -6,38 +6,35 @@ interface User {
...
   constructor() {
-    this.users = []
-    this.currentId = 1
+    this._users = []
   }
 
   private assignId() {
-    this.currentId++
-    return this.currentId - 1
+    return Date.now()
```
DuHerb: if I remember correctly, `Date.now()` returns a number.  Best practice would be to convert this as a string.

DuHerb: there is a method you can chain to the end of your `Date.now()` invocation.

src/index.ts
```diff
@@ -46,15 +43,40 @@ class UserAPI {
   }
 
   getUsers() {
-
+    if ((!this._users) || Array.isArray(this._users) === false) {
```
DuHerb: can be simplified to `if (!this.users || !Array.isArray(this.users))` or even simpler `if (!this.users)`

DuHerb: the first case will be more appropriate for JS, as you are enforcing that the object is an array.  This is good.

when we convert to TS, you will see that it is redundant and the second `if` is better

dlindsey-olioapps: thanks! this is updated to be just if (!this.users)   -- at least till TypeScript

src/index.ts
dlindsey-olioapps: thanks! this is to.String() now

src/index.ts
```diff
@@ -2,42 +2,54 @@ interface User {
...
 
   constructor() {
-    this.users = []
-    this.currentId = 1
-  }
-
-  private assignId() {
-    this.currentId++
-    return this.currentId - 1
+    this._users = [
```
DuHerb: same comment as made to Andy.

`UserApi` constructor should accept an optional parameter of `Array<User>`.  If provided, the constructor method sets `this.users` as that value.  Otherwise, the list is empty.

When testing, you create your mockUsers list and pass that to the `UserApi` when creating it.

dlindsey-olioapps: Thanks, that makes sense! Should be fixed

DuHerb (approved):

PR #34, opened 2022-06-08
Minoka's eventstream puzzler draft

kakizaki55:
**1. Ticket:**
Issue #none

**2. What this does:**
* event stream puzzler function written 
* extrapolated function into 2 more smaller functions 
* test for all function, all test passing. 
 
**3. Media:**

Screen Shot 2022-06-08 at 3 58 16 PM

**4. How to test:**
pull down code and run ```yarn test```


src/index.test.ts
```diff
@@ -1,7 +1,61 @@
...
+  firstMockEventStream,
+  secondMockEventStream,
+  thirdMockEventStream,
+  fourthMockEventStream
+} from './mockEventStream'
+import { getHighestScoringRegion } from "./index"
+
 
 describe('Tests will go here!', () => {
   it('should pass', () => {
```
DuHerb: remove this dummy test

src/index.test.ts
```diff
@@ -1,7 +1,61 @@
...
+      { timestamp: 123123125, eventType: 'view' },
+      { timestamp: 123123124, eventType: 'screenshot' },
+      { timestamp: 123123125, eventType: 'new message' },
+      { timestamp: 123123125, eventType: 'view' },
+      { timestamp: 123123125, eventType: 'screenshot' },
+    ]
+    expect(highestScoringRegion).toEqual(expected)
+  })
+
 })
```
DuHerb: new line error

src/index.test.ts
```diff
@@ -1,7 +1,61 @@
...
+} from './mockEventStream'
+import { getHighestScoringRegion } from "./index"
+
 
 describe('Tests will go here!', () => {
   it('should pass', () => {
     expect(true).toBeTruthy()
   })
+
+  it('should return the correct array region', () => {
```
DuHerb: all of these `it` descriptions say the same thing.  can you provide more context as to what these are testing?

If you are just feeding different streams to show that they are correctly being scored, then multiple expects can go into the same `it` block.  And they should all pass.

I see one that is returning only three events.  That block is testing a different functionality.  So it merits it's own `it` (or `describe, based on the complexity of the testing`), as well as a unique description.

src/index.test.ts
```diff
@@ -1,7 +1,82 @@
...
+  fourthMockEventStream,
+  firstMockRegion,
+  secondMockRegion,
+} from './mockEventStream'
+import { getHighestScoringRegion, findRegionScore, scoreEvent } from "./index"
 
-describe('Tests will go here!', () => {
-  it('should pass', () => {
-    expect(true).toBeTruthy()
+describe('testing all functions for getting the highest scoring region of event stream', () => {
```
DuHerb: great.

when organizing tests, use `describe` blocks to group tests together in logical structures. 
Here you may have a top level describe called 'ScoreEventStream` to describe the suite.
Then you nest a `describe(scoreEvent...)` with each `if` block under there.
then `describe('findRegionScore'...) ......
and so on.

src/index.test.ts
```diff
@@ -1,7 +1,91 @@
...
+  secondMockRegion,
+} from './mockEventStream'
+import { getHighestScoringRegion, findRegionScore, scoreEvent } from "./index"
 
-describe('Tests will go here!', () => {
-  it('should pass', () => {
-    expect(true).toBeTruthy()
+describe('Event scoring function tests.', () => {
+  
+  describe('Tests for getting the highest scoring region in a event stream', () => {
```
DuHerb: describe text should just be the function being tested.  `getHighestScoringRegions`

src/index.test.ts
```diff
@@ -1,7 +1,91 @@
...
+        { timestamp: 123123124, eventType: 'screenshot' },
+        { timestamp: 123123125, eventType: 'new message' },
+        { timestamp: 123123125, eventType: 'view' },
+        { timestamp: 123123125, eventType: 'screenshot' },
+      ]
+      expect(highestScoringRegion).toEqual(expected)
+    })
+  })
+
+  describe('Tests for getting the total score of each region', () => {
```
DuHerb: same.  describe string should be function name.

src/index.test.ts
DuHerb: the `it` blocks describe exactly how the described function is being tested in each case.

src/index.test.ts
```diff
@@ -1,7 +1,91 @@
...
+        [
+          { timestamp: 123123125, eventType: 'new message' },
+          { timestamp: 123123125, eventType: 'view' },
+          { timestamp: 123123125, eventType: 'screenshot' },
+        ])
+        expect(thirdRegionScore).toEqual(6)
+    })
+  })
+  
+  describe('Test for getting the correct score for each event', () => {
```
DuHerb: same

DuHerb (approved):

PR #36, opened 2022-06-09
Add jest tests for all methods, add immutability to methods

dlindsey-olioapps:
**Ticket**
Not applicable

**What does this PR do**
- Immutability has been added to all methods
- Writes Jest tests for all methods
- Removes original console.log tests

**Media**
image

**How to test**
To test, clone the repository to your computer and navigate to the project with the terminal. Download all necessary packages with **_yarn install_** and when ready, run **_yarn run test_** to execute the application. You should see several tests and the results of these tests.


src/index.test.ts
```diff
@@ -1,7 +1,188 @@
-//write tests here
+import { UserAPI } from "./index"
 
-describe('Tests will go here!', () => {
-  it('should pass', () => {
-    expect(true).toBeTruthy()
+describe('new UserAPI()', () => {
```
DuHerb: organizational nit pick (just my own preference, no change required).
I would nest ALL subsequent `describe` blocks in this one.  (describes can be nested inside each other)

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+
+  const x = new UserAPI()
+
+  it("should confirm x is a UserAPI object", () => {
```
DuHerb: this test should be wrapped in it's own `describe` as something like `describe("UserApi constructor"...)...`.  Then you have tests for:

- do you get an instance of UserApi (as seen here)?
- when called with no argument, does get users return an empty list?
- when called with an argument (list of users), does getUsers immediately return the expected list?

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+    expect(x).toBeInstanceOf(UserAPI)
+  })
+})
+
+describe('UserAPI.addUser()', () => {
+
+  it("should return the new user", () => {
+    const x = new UserAPI([{ name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" }])
```
DuHerb: avoid naming variable arbitrary letters.  variable names should describe the thing.

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+    jest.spyOn(Date, "now").mockReturnValue(new Date(1587893830000).getTime())
+
+    expect(x.addUser({ name: 'Daniel', favoriteColor: 'purple', age: 33 }))
```
DuHerb: for clarity, Olio generally prefers if `expected` and `actual` arguments are defined as constants, rather than have values passed directly into the `expect` and `toEqual` functions.

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+      .toEqual({ name: 'Daniel', favoriteColor: 'purple', age: 33, id: "1587893830000" })
+    expect(x.getUsers())
```
DuHerb: this test is doing something other than what is described.  Should be wrapped in it's own `it`, or moved into the the `getUsers` describe block

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+
+  it("should delete one user and return deleted user", () => {
+    const x = new UserAPI([
+      { name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" },
+      { name: 'Daniel', favoriteColor: 'green', age: 33, id: "2" },
+      { name: 'Jenny', favoriteColor: 'yellow', age: 88, id: "3" }
+    ])
+    expect(x.deleteUserById('1'))
+      .toEqual({ name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" })
+    expect(x.getUsers())
```
DuHerb: again, should be in it's own test block, as it is fundamentally different test from 'delete user returns deleted user`

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+      { name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" },
+      { name: 'Daniel', favoriteColor: 'green', age: 33, id: "2" },
+      { name: 'Jenny', favoriteColor: 'yellow', age: 88, id: "3" },
+      { name: 'Daniel', favoriteColor: 'red', age: 48, id: "4" }
+    ])
+    expect(x.searchUserByName("DaNIeL")).toEqual([
+      { name: 'Daniel', favoriteColor: 'green', age: 33, id: "2" },
+      { name: 'Daniel', favoriteColor: 'red', age: 48, id: "4" }
+    ])
+    expect(x.getUsers())
```
DuHerb: unnecessary

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+  })
+
+  it("should fail to find user with name Billy", () => {
+    const x = new UserAPI([
+      { name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" },
+      { name: 'Daniel', favoriteColor: 'red', age: 48, id: "2" }
+    ])
+    expect(() => { x.searchUserByName("Billy") })
+      .toThrow("User(s) not found")
+    expect(x.getUsers())
```
DuHerb: unnecessary.  this is covered in a different test

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+      { name: 'Jenny', favoriteColor: 'yellow', age: 88, id: "3" },
+      { name: 'Daniel', favoriteColor: 'red', age: 48, id: "4" }
+    ])
+    expect(x.searchUsersByFavoriteColor("YeLlOw"))
+      .toEqual([
+        { name: 'Daniel', favoriteColor: 'yellow', age: 33, id: "2" },
+        { name: 'Jenny', favoriteColor: 'yellow', age: 88, id: "3" },
+      ])
+    expect(x.getUsers())
+      .toEqual([
```
DuHerb: same

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+  })
+
+  it("should fail to find user with favorite color orange", () => {
+    const x = new UserAPI([
+      { name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" },
+      { name: 'Daniel', favoriteColor: 'yellow', age: 33, id: "2" }
+    ])
+    expect(() => { x.searchUsersByFavoriteColor("Orange") })
+      .toThrow("User(s) not found")
+    expect(x.getUsers())
```
DuHerb: same

DuHerb (changes requested):

src/index.test.ts
```diff
@@ -1,7 +1,188 @@
...
+      .toEqual([])
+  })
+
+  it("should fail to add user if user with same props exists", () => {
+    const x = new UserAPI([{ name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" }])
+    const userDuplicate = { name: 'Larry', favoriteColor: 'gray', age: 544 }
+    expect(() => { x.addUser(userDuplicate) })
+      .toThrow("User with these properties already exists")
+    expect(x.getUsers())
```
DuHerb: testing the exception, not getUsers

src/index.test.ts
dlindsey-olioapps: Thanks! Just to be sure, you're referring to jest.spyOn right? That is fixed in the new code with objects. For the above - I wasn't sure how to get it to work with the arrays, as the ID was being generated was with Date.now(), and would return an object with an ID that I wouldn't know before making the test.

DuHerb (changes requested):

src/index.test.ts
dlindsey-olioapps: Thanks! I've got them removed now. I figured I would be extra cautious and confirm that it was actually removing/adding/etc it from the array. 

Is there a correct way to test that the array has been correctly modified? Would I just move my _new userAPI()_ instantiation outside of the it()? Or should I just be avoiding it all together?

src/index.test.ts
```diff
@@ -1,7 +1,160 @@
...
+
+  describe('UserAPI.deleteUserById()', () => {
+
+    const users = new UserAPI([
+      { name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" },
+      { name: 'Daniel', favoriteColor: 'green', age: 33, id: "2" },
+      { name: 'Jenny', favoriteColor: 'yellow', age: 88, id: "3" }
+    ])
+
+    it("should delete one user and return deleted user", () => {
```
DuHerb: should also expect the returned use to equal `{ name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" }`

src/index.test.ts
```diff
@@ -1,7 +1,160 @@
...
+  describe('UserAPI.searchUserByName()', () => {
+
+    const users = new UserAPI([
+      { name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" },
+      { name: 'Daniel', favoriteColor: 'green', age: 33, id: "2" },
+      { name: 'Jenny', favoriteColor: 'yellow', age: 88, id: "3" },
+      { name: 'Daniel', favoriteColor: 'red', age: 48, id: "4" }
+    ])
+
+    it("should return array of two DaNIeL users (ignore letter case)", () => {
```
DuHerb: change not necessary, but the descriptions should be slightly more generic....ie `search ignores letter case`

src/index.test.ts
```diff
@@ -1,7 +1,160 @@
...
+    const users = new UserAPI([
+      { name: 'Larry', favoriteColor: 'gray', age: 544, id: "1" },
+      { name: 'Daniel', favoriteColor: 'yellow', age: 33, id: "2" },
+      { name: 'Jenny', favoriteColor: 'yellow', age: 88, id: "3" },
+      { name: 'Daniel', favoriteColor: 'red', age: 48, id: "4" }
+    ])
+
+    it("should return array of two same fav color users (ignore letter case)", () => {
+      expect(users.searchUsersByFavoriteColor("YeLlOw"))
+        .toEqual([
```
DuHerb: Olio convention would be to define `actual` and `expected` constants.  this is strictly for readability/clarity.

so your testing statements would look like:
`expect(actual).toEqual(expected)`

Exceptions can be made for simple values like booleans, numbers, or short strings.
The way you've done your expections testing is also fine.

For the sake of this exercise, changes aren't necessary.  But future tests should consider this pattern.

DuHerb (changes requested):

DuHerb (approved):

PR #38, opened 2022-06-10
Convert array dataset to object structure

Plymatea:
**Ticket**
None

**This Does**
- Refcator index.ts dataset to use object storage instead of array storage
- `getUsers()` still returns an array of users
- Refactor index.test.ts for updated dataset structure
- Refactor `deleteUserById()` to return object of user in stead of array
- Add explicit return types to `UserAPI` methods

**Media**
- Shows all tests passing

image

**Test**
none


src/index.test.ts
```diff
@@ -63,10 +62,11 @@ describe(UserAPI, () => {
       const userAPI = new UserAPI(seedUsers)
       userAPI.deleteUserById("1")
       const actual = userAPI.getUsers()
-      const expected = [
-        { id: "2", name: "Andy",  favColor: "purple", age: 150 },
-        { id: "3", name: "Sarah", favColor: "Blue", age: 200}
-      ]
+      const expected = [ seedUsers['2'], seedUsers['3'] ]
+      console.log("actual", actual);
+      console.log("expected", expected);
```
DuHerb: kill logs

src/index.test.ts
```diff
@@ -1,12 +1,11 @@
...
-  { id: "1", name: "andy", favColor: "blue", age: 247 },
-  { id: "2", name: "Andy",  favColor: "purple", age: 150 },
-  { id: "3", name: "Sarah", favColor: "Blue", age: 200} 
-]
+const seedUsers: Record<string, IUser> = {
+  "1": { id: "1", name: "andy", favColor: "blue", age: 247 },
+  "2": { id: "2", name: "Andy",  favColor: "purple", age: 150 },
+  "3": { id: "3", name: "Sarah", favColor: "Blue", age: 200} 
+}
 const userAPI = new UserAPI(seedUsers)
```
DuHerb: userApi should be initiated in each different describe block.  Here, your userApi will always have users, which could possibly pollute/affect tests in unexpected ways.

DuHerb: sorry that i missed this previously.

src/index.ts
```diff
@@ -6,31 +6,36 @@ export interface IUser {
...
   } 
 
-  addUser(user: IUser) {
-    if (user.id) {
+  addUser(userToAdd: IUser) {
+    if (userToAdd.id) {
       throw new SyntaxError("Please resubmit without pre-exisitng ID field.")
     } 
+    let user = {...userToAdd}
+    const result = (Object.values(this._users)).filter( existingUser => (
```
DuHerb: small optimization....
instead of `.filter`, could just use `.some`, returning a boolean.  Or, if you actually need the found item for something, '.find'

src/index.ts
```diff
@@ -6,31 +6,36 @@ export interface IUser {
...
-    user.id = Date.now().toString()
-    const result = (this._users).find( ( { name, favColor, age }: IUser) => (name === user.name && favColor === user.favColor && age === user.age))
-    
-    if (result) {
+    if (result.length) {
       throw new Error("A user with those properties already exists in the database.")
     } 
-    
-    this._users = [...this._users, user]
+    const id = Date.now().toString() + Math.floor(Math.random()*100)
```
DuHerb: consider making a private `getUId` function that provides this value for you.

just a small example of encapsulation.

src/index.ts
```diff
@@ -6,31 +6,36 @@ export interface IUser {
...
+    user.id = id
+    let newUsersState: any = {...this._users}
```
DuHerb: `let` ==> `const`, and the newUser can be added to `newUserState` using the spread method, rather than mutating the object (`ln32`)

src/index.ts
```diff
@@ -6,31 +6,36 @@ export interface IUser {
...
+    newUsersState[id] = user
+    this._users = {...newUsersState}
     return user
   }
 
   getUserById(id: string) {
-    const foundUser = this._users.find( user => user.id === id) 
-    
+    const foundUser = Object.values(this._users).find( user => user.id === id)
```
DuHerb: now that were using the object, can just get the user by the id `this._users[id]`.  this is one of the advantages of using an object (or map) for collections.

src/index.ts
```diff
@@ -41,23 +46,24 @@ export class UserAPI {
...
   deleteUserById(id: string) {
-    const user = this._users.filter( user => user.id === id)
+    const user = this._users[id]
     
-    if (user.length) {
-      const newUserState = this._users.filter( user => user.id != id)
-      this._users = [...newUserState]
+    if (user) {
+      const newUserState = {...this._users}
+      delete newUserState[id]
```
DuHerb: even though we are mutating an object we just created, we can still strictly adhere to immutably principles here by using the spread operator.  try using `...rest`, and let me know if you get stuck after more than 15/20 minutes.

src/index.ts
```diff
@@ -66,10 +72,20 @@ export class UserAPI {
...
+    const users = Object.values(this._users).filter( user => user.favColor.toLowerCase() === color.toLowerCase())
     if (users.length) {
       return users
     } 
     throw new ReferenceError(`No users favorite color is ${color}`)
   }
 }
+
+
+const seedUsers: Record<string, IUser> = {
```
DuHerb: if this was for dev....it shouldn't be included in PR

DuHerb (changes requested):

src/index.test.ts
Plymatea: Damn Sorry. Done

src/index.test.ts
Plymatea: Done

src/index.ts

src/index.ts
Plymatea: Good call! Done

src/index.ts
Plymatea: Noticed this after i created the PR. I must have missed pushing the changes back up to github.

src/index.ts
Plymatea: nice. changed this.

src/index.ts

src/index.ts
Plymatea: Thank you For the help on this!  Done

DuHerb (approved):

PR #39, opened 2022-06-13
TS data verification

Plymatea:
**Ticket:**
- None

**What this Does:**
- Adds Read Only verification to tests, methods, and properties
- Files should all ready have TS verification for method parameters, and output. 

**Media:**
None

**Testing**
- All jest test still pass


.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
+{
```
DuHerb, line 1: this `launch.json` file should be listed in your `.gitignore` file

src/index.ts
```diff
@@ -1 +1,78 @@
-//Define class here
\ No newline at end of file
+export interface IUser {
+  readonly id?: string
+  name: string
+  favColor: string
+  age: number
```
DuHerb: all properties here should be `readonly`

src/index.ts
```diff
@@ -1 +1,78 @@
...
+}
+
+export class UserAPI {
+  private _users: Readonly<Record<string, IUser>>
+
+  constructor(seedData: Record<string, IUser> = {}) {
```
DuHerb: Wouldn't be a bad thing to define/export an `IUsers` type. `type IUser = Record<string, IUser>`

only suggested here because we use this list as a type in several places.

src/index.ts
```diff
@@ -1 +1,78 @@
...
+    this._users = {...seedData}
+  } 
+
+  private _assignId(): string {
+    const id: Readonly<string> = Date.now().toString() + Math.floor(Math.random()*100)
+    return id
+  }
+  
+  addUser(user: IUser): IUser{
```
```suggestion
  addUser(user: IUser): IUser {
```

space ^^

DuHerb: looks like space is missing after return types on several methods.  should fix

src/index.ts
```diff
@@ -1 +1,78 @@
...
+  deleteUserById(id: string): IUser {
+    if (this._users[id]) {
+      const { [id]: deletedUser, ...rest } = this._users
+      this._users = {...rest}
+      return deletedUser
+    } 
+    throw new ReferenceError(`No user found with id ${id}.`)
+  }
+
+  searchUserByName(name: string): Array<IUser>{
```
DuHerb: `ReadonlyArray`

src/index.ts
```diff
@@ -1 +1,78 @@
...
+    const users = Object.values(this._users).filter( user => user.name.toLowerCase() === name.toLowerCase())
+    
+    if (users.length) {
+      return users
+    } 
+    throw new ReferenceError(`No user found with name ${name}`)
+  }
+
+  searchUsersByFavoriteColor(color: string): Array<IUser>{
```
DuHerb: `ReadonlyArray`

DuHerb (changes requested):

.vscode/launch.json
Plymatea: Done,  I don't think I understand git versioning well enough to scrub the file from the git repo history.  I did it once on a personal project, but scrubbed the entire repo. I'm afraid I'll screw something up if I try it here. 

I did add the /vscode folder the gitignore though.

src/index.ts

src/index.ts

src/index.ts

src/index.ts

src/index.ts

PR #40, opened 2022-06-14
Added code and tests for event stream puzzler

dlindsey-olioapps:
**Ticket:**
N/a

**What this PR does**
This PR request adds the following
- Planning document with notes
- Converts class and methods code into single function, and two interfaces
- Jest for testing function

**Media**
image

**How to test**
To test, clone the repository to your computer and navigate to the project with the terminal. Download all necessary packages with **_yarn install_** and when ready, run **_yarn run test_** to execute the application. You should see several tests and the results of these tests.

To use your own seed data, please enter the following at the bottom of index.ts, replacing [] with your event stream input array.

```
const seedData: ReadonlyArray<Event> = []
console.log(getWinningRegion(seedData))
```
Once you've added in your code, in the terminal run **_yarn run dev_** to see the output

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -2,7 +2,7 @@
 import { EventStream, EventInput, Region } from './index'
 
 describe('Class EventStream', () => {
-  const seedData: Array<EventInput> = [
+  const seedData: ReadonlyArray<EventInput> = [
     {
       timestamp: 123123123,
       eventType: "newMessage"
@@ -40,7 +40,7 @@ describe('Class EventStream', () => {
       eventType: "newMessage"
     }
   ]
-  const largerSeedData: Array<EventInput> = [
+  const largerSeedData: ReadonlyArray<EventInput> = [
     {
       timestamp: 123123123,
       eventType: "newMessage"
@@ -117,19 +117,19 @@ describe('Class EventStream', () => {
   describe('Class Instantiation of EventStream', () => {
 
     it("should confirm eventStream is a EventStream object", () => {
-      const eventStream = new EventStream(seedData)
+      const eventStream = new EventStream([...seedData])
       expect(eventStream).toBeInstanceOf(EventStream)
     })
 
     it("should confirm events get added to eventStream object", () => {
-      const eventStream = new EventStream(seedData)
-      const actual: Array<EventInput> = eventStream.getUnsortedScores()
-      const expected: Array<EventInput> = seedData
+      const eventStream = new EventStream([...seedData])
+      const actual: ReadonlyArray<EventInput> = eventStream.getUnsortedScores()
+      const expected: ReadonlyArray<EventInput> = seedData
       expect(expected).toEqual(actual)
     })
 
     it("should fail with not enough obj provided", () => {
-      const badSeedData: Array<EventInput> = [
+      const badSeedData: ReadonlyArray<EventInput> = [
         {
           timestamp: 123123123,
           eventType: "newMessage"
@@ -147,12 +147,12 @@ describe('Class EventStream', () => {
           eventType: "newMessage"
         }
       ]
-      const badEventStream: {} = () => new EventStream(badSeedData)
+      const badEventStream: {} = () => new EventStream([...badSeedData])
       expect(badEventStream).toThrow("Not enough events to score. Please provide at least five")
     })
 
     it("should fail with if invalid event type provided", () => {
-      const badSeedData: Array<EventInput> = [
+      const badSeedData: ReadonlyArray<EventInput> = [
         {
           timestamp: 123123123,
           eventType: "newMessage"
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1 +1,78 @@
-//Define class/functions here
\ No newline at end of file
+export interface EventInput {
```
```suggestion
export interface Event {
```

src/index.ts
```diff
@@ -1 +1,78 @@
...
+  timestamp: Readonly<number>,
+  eventType: Readonly<string>
```
DuHerb: this type should limit values to only be one of the three defined options.

src/index.ts
```diff
@@ -1 +1,78 @@
...
+export interface Region {
+  regionId: Readonly<number>,
+  score: Readonly<number>,
+  inputLocations: ReadonlyArray<number>
+}
+
+export class EventStream {
+
+  private _scoreTable: Record<Readonly<string>, Readonly<number>> = {
+    ["newMessage"]: 1,
```
DuHerb: brackets shouldn't be needed here

src/index.ts
```diff
@@ -1 +1,78 @@
...
+    }
+    this._setSortedWinners()
+  }
+
+  getUnsortedEvents(): ReadonlyArray<EventInput> {
+    return [...this._data]
+  }
+
+  private _setSortedWinners() {
+    const numOfRegions: number = 1 + (this._data.length - 5)
```
DuHerb: don't have to type this as number.  easily inferred by TS

src/index.ts
```diff
@@ -1 +1,78 @@
...
+    if (numOfRegions < 1) {
+      throw "Not enough events to score. Please provide at least five"
```
DuHerb: if the number of events is less than or equal to five, the function should return all events.

src/index.test.ts
```diff
@@ -1,7 +1,303 @@
...
+      expect(expected).toEqual(actual)
+    })
+
+    it("should fail to create EventStream with wrong input data type", () => {
+      // const eventStream = new EventStream({...seedData})
+      const badEventStream: {} = () => new EventStream({ ...seedData })
+      expect(badEventStream).toThrow("Input of object was found. Please provide a valid array.")
+    })
+
+    it("should fail with not enough obj provided", () => {
```
DuHerb: if number of events is less than or equal to five, return all items.

src/index.test.ts
```diff
@@ -1,7 +1,303 @@
...
+          timestamp: 123123125,
+          eventType: "newMessage"
+        }
+      ]
+      const badEventStream: {} = () => new EventStream([...badSeedData])
+      expect(badEventStream).toThrow(`Invalid event type found (event: 4, timestamp: 123123125, type: porcupine)`)
+    })
+  })
+
+  describe('getUnsortedScores()', () => {
```
DuHerb: this is not a requirement of the puzzle.

src/index.test.ts
```diff
@@ -1,7 +1,303 @@
...
+          eventType: "screenshot"
+        }
+      ]
+      const eventStream = new EventStream(largerSeedData)
+      const actual: Array<EventInput> = eventStream.getHighestScoringRegion()
+      expect(expected).toEqual(actual)
+    })
+  })
+
+  describe('getSortedScores()', () => {
```
DuHerb: not part of the requirments

src/index.ts
```diff
@@ -1 +1,78 @@
...
+      if (!this._scoreTable[eventType]) {
+        throw `Invalid event type found (event: ${y}, timestamp: ${this._data[y].timestamp}, type: ${eventType})`
+      }
+      regionNumbers.push(y)
+      regionScore += this._scoreTable[eventType]
+    }
+    return {
+      regionId: regionNum,
+      score: regionScore,
+      inputLocations: [...regionNumbers]
```
DuHerb: i'm unclear about the purpose of inputLocations.  If we know the starting index of the highest scoring region, then we can calculate the range of the region.

DuHerb (changes requested):

DuHerb:
Try to refactor all of your for loops to use Higher Order functions like `reduce` or `map`

src/index.ts
dlindsey-olioapps: Thanks! It was for getSortedScores() but I will remove that, then the input locations, per your previous comment

src/index.ts
```diff
@@ -1 +1,32 @@
-//Define class/functions here
\ No newline at end of file
+export interface Event {
+  timestamp: Readonly<number>,
+  eventType: Readonly<"newMessage" | "view" | "screenshot">
```
DuHerb, line 3: using the `readonly` decorator would also be fine.  

```
...
readonly timestamp: number,
...
```

src/index.ts
```diff
@@ -1 +1,32 @@
...
+}
+
+interface Region {
+  regionStartId: Readonly<number>,
+  score: Readonly<number>
+}
+
+export function getWinningRegion(inputData: ReadonlyArray<Event>): ReadonlyArray<Event> {
+  const scoreTable = {
```
DuHerb, line 12: for the sake of this excercise, i'd like you:
encapsulate the scoring of a single event into a single function. Should return a number
encapsulate the scoring of an array into a single function. Should return a number

DuHerb (changes requested):

src/index.ts
dlindsey-olioapps: Thanks! Hopefully my changes are what you were looking for!

PR #41, opened 2022-06-14
Daniel user api

DuHerb:
DO NOT MERGE
For review only.


src/index.ts
```diff
@@ -1 +1,73 @@
-//Define class here
\ No newline at end of file
+export interface IUser {
+  name: string
+  favoriteColor: string
+  age: number
```
DuHerb, line 4: all properties here should be `readonly`

src/index.ts
```diff
@@ -1 +1,73 @@
...
+  readonly id?: string
+}
+
+export class UserAPI {
+
+  constructor(private _users: Record<string, IUser> = {}) {
```
DuHerb, line 10: an exported `Users` type would be beneficial here, since it is a domain shape that is used in many places.  `export type Users = Record<string, IUser>`

src/index.ts
```diff
@@ -1 +1,73 @@
...
+        x.favoriteColor.toLowerCase() === x.favoriteColor.toLowerCase() &&
+        x.age === user.age
+    )) {
+      throw new Error("User with these properties already exists")
+    }
+    this._users = { ...this._users, [id]: { ...user, id: id } }
+    return { ...user, id: id }
+  }
+
+  getUserById(id: string): Readonly<IUser> {
```
DuHerb, line 30:
```suggestion
  getUserById(id: string): IUser {
```

because `IUser` is readonly (or will be once all of its properties are `readonly`), we don't need `Readonly` here.

there is a pattern you might see in some places where a declared interface extends an empty Readonly object, so that the properties don't have to be individually declared as `readonly`.  That pattern would also be acceptable

PR #42, opened 2022-06-14
Solve Event puzzler with tests

Plymatea:
**Ticket:**
- None

**What this Does:**
- Solves the event puzzler
- Creates tests for solution
- Adds new file for mock testing data

**Media:**
image

**Testing**
- None

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,40 +6,42 @@ export interface IEvent {
 export type eventTypeState = 'newMessage' | 'screenShot' | 'view'
 
 export function eventStream(stream: Array<IEvent>): ReadonlyArray<IEvent> {
-  if(stream.length <= 5 ) { return stream }
+  if (stream.length <= 5) return stream
 
-	const newStream: Array<IEvent> = stream.map( event => ({...event, value: assignPointValue(event.eventType)}))
+  const newStream: Array<IEvent> = stream.map((event) => ({
+    ...event,
+    value: assignPointValue(event.eventType),
+  }))
   const index: number = findLastIndexOfRegionWithHighestSum(newStream)
-	const region: ReadonlyArray<IEvent> = stream.splice((index-4), 5) 
-	return region
+  const region: ReadonlyArray<IEvent> = stream.splice(index - 4, 5)
+  return region
 }
 
 export function assignPointValue(eventType: eventTypeState): Readonly<number> {
-	switch (eventType) {
-		case 'newMessage':
-			return 1
-		case 'view':
-			return 2
-		case 'screenShot':
-			return 3
-		default:
-			throw new Error("Not a valid eventType")
-	}
+  switch (eventType) {
+    case 'newMessage':
+      return 1
+    case 'view':
+      return 2
+    case 'screenShot':
+      return 3
+    default:
+      throw new Error('Not a valid eventType')
+  }
 }
 
 export function findLastIndexOfRegionWithHighestSum(stream: Array<IEvent>): number {
-  if(stream.length <= 5 ) { return stream.length-1 }
+  if (stream.length <= 5) return stream.length - 1
 
   let index = 4
   let maxCumulativeSum = 0
-  for(let i=4; i<stream.length; i++) {
-    const currentSum = (
-      stream[i].value! + 
-      stream[i-1].value! + 
-      stream[i-2].value! + 
-      stream[i-3].value! + 
-      stream[i-4].value!
-    )
+  for (let i = 4; i < stream.length; i++) {
+    const currentSum =
+      stream[i].value! +
+      stream[i - 1].value! +
+      stream[i - 2].value! +
+      stream[i - 3].value! +
+      stream[i - 4].value!
     if (currentSum > maxCumulativeSum) {
       maxCumulativeSum = currentSum
       index = i
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,23 +1,24 @@
 export interface IEvent {
-  readonly timestamp: number,
-  readonly eventType: eventTypeState,
-  value?: number
+  readonly timestamp: number
+  readonly eventType: EventType
 }
-export type eventTypeState = 'newMessage' | 'screenShot' | 'view'
+export type EventType = 'newMessage' | 'screenShot' | 'view'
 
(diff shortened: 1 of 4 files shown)


src/mockTestData.ts
```diff
@@ -0,0 +1,148 @@
...
+	},
+	{
+		timestamp: 123123125,
+		eventType: 'view'
+	},
+	{
+		timestamp: 123123125,
+		eventType: 'view'
+	}
+]
```
DuHerb: new line.  

Also, indentation is all over the place in this file, and between all files.  Tab should be the same in all files.  In general, I would go with 2.

src/index.ts
```diff
@@ -1 +1,41 @@
...
+export interface IEvent {
+  readonly timestamp: number,
+  readonly eventType: eventTypeState,
+}
+export type eventTypeState = 'newMessage' | 'screenShot' | 'view'
+
+export function eventStream(stream: ReadonlyArray<IEvent>): ReadonlyArray<IEvent> {
+  if(stream.length <= 5 ) { return stream }
+
+	const newStream: Array<IEvent> =  [...stream]
```
DuHerb: shouldn't need a `newStream` or spread, as you are only going to iterate over the input stream.

src/index.ts
```diff
@@ -1 +1,41 @@
...
+	const newStreamValues: Array<number>= []
+	stream.forEach( e => newStreamValues.push(assignPointValue(e.eventType)));
```
DuHerb: can you refactor this to use `map` or `reduce`

DuHerb (changes requested):

src/index.ts
```diff
@@ -1 +1,41 @@
...
+	
+	let index = 4 //Start with the 5th event in the stream becuase the for loop looks backward and ends at the end of the array
+	let highestValue = 0
+	for(let i=4; i<newStreamValues.length; i++) {  // Same here, Want to start at the 5th event in the array
```
DuHerb: i'm not sure I understand why we need to look backwards over results.

In general, I'm finding the algorithm difficult to understand.  One way to mitigate readability/maintainability, is to further encapsulate logic into smaller functions.  

You may find it useful to build a helper function that takes an array of events and just provides a score.  Or similar such function.

DuHerb (changes requested):

src/mockTestData.ts
Plymatea: I'm not totally sure why the indentation is all over the place. My standard is 2 spaces for a tab. I did find three spot on my local branch that I need to tab in.  But my local branch is not nearly as screwy as this.... I'll pay more attention after I push it up this next time.

src/index.ts
Plymatea: done

src/index.ts
Plymatea: I encapsulated finding the index into a helper function. 

I managed to get rid of the nested for loop, but not really happy with the hard coding in finding the sum of the values.  Is there a better way to sum the values of a sub array?

src/index.ts
Plymatea: done

src/index.ts
```diff
@@ -1 +1,51 @@
-//Define class/functions here
\ No newline at end of file
+export interface IEvent {
+  readonly timestamp: number,
+  readonly eventType: eventTypeState,
+  value?: number
+}
+export type eventTypeState = 'newMessage' | 'screenShot' | 'view'
```
```suggestion
export type EventType = 'newMessage' | 'screenShot' | 'view'
```

types/interfaces should be capitalized

src/index.ts
```diff
@@ -1 +1,51 @@
...
+}
+
+export function findLastIndexOfRegionWithHighestSum(stream: Array<IEvent>): number {
+  if (stream.length <= 5) return stream.length - 1
+
+  let index = 4
+  let maxCumulativeSum = 0
+  for (let i = 4; i < stream.length; i++) {
+    const currentSum =
+      stream[i].value! +
```
DuHerb: adding the `value` prop and then looking it up feels like an unnecessary step.  Can just get the value directly here via `assignPointValue(stream[i].eventType)`

DuHerb:
In general, I like that you've encapsulated different logic in to smaller, more maintainable code.

however, the fact that we are still looking backwards into arrays smells fishy to me.

can you think of a solution where your function can identify/record the index of the first event in the high scoring region?

src/index.ts
```diff
@@ -1 +1,51 @@
-//Define class/functions here
\ No newline at end of file
+export interface IEvent {
+  readonly timestamp: number,
+  readonly eventType: eventTypeState,
+  value?: number
```
DuHerb: extra step.  see comment in `findLastIndex....`

src/index.ts
```diff
@@ -1 +1,51 @@
...
+  }
+}
+
+export function findLastIndexOfRegionWithHighestSum(stream: Array<IEvent>): number {
+  if (stream.length <= 5) return stream.length - 1
+
+  let index = 4
+  let maxCumulativeSum = 0
+  for (let i = 4; i < stream.length; i++) {
+    const currentSum =
```
DuHerb: getting the scores for an array of events can be delegated to a helper function.

The current solution, while working for this exact use case, is not very maintainable.  For instance, what if the requirement for the function changed so that the length of the scored array was changed to 7.  Now you have to add two more lines here, and adjust hardcoded values all over the place.

In another scenario, what if the the requirement was that the region length was completely variable?  In that case, this solution won't work at all.

Encapsulation helps handle those types of solutions.

DuHerb (changes requested):

src/index.ts

Plymatea:
> In general, I like that you've encapsulated different logic in to smaller, more maintainable code.
> 
> however, the fact that we are still looking backwards into arrays smells fishy to me.
> 
> can you think of a solution where your function can identify/record the index of the first event in the high scoring region?

Instead of assigning the value to each object, I decided to assign the region value to each object. Completed that and realized it was just easier to create an array of the region values. So I refactored it again to it's current state. Create an array of values corresponding to the regions.  Then find the max value, and index, then return the region from the original stream. 

This seems a lot nicer than my first approach. Thank you.

PR #47, opened 2023-08-15
Emma tdd lesson1 fail tests

emmalgs:
## What this is
Red phase for Lesson 1 TDD
* Write failing tests in Typescript for Lesson 1 TDD workshop

## Ticket
n/a

## Steps to Verify
* run ```yarn test:watch``` in project directory on 'emma-tdd-lesson1-fail-tests' branch

## Media Section
_Failing tests_
Screenshot 2023-08-16 at 9 47 53 AM

PR #52, opened 2023-08-16
Emma tdd lesson1 pass tests

emmalgs:
## What This Is
Green Phase for Lesson 1 Testing
* All tests passing for Lesson 1

## Ticket
n/a

## Steps To Verify
* In the project directory on branch 'emma-tdd-lesson1-pass-tests' run ```yarn test```

## Media
_Passing tests in terminal_
Screenshot 2023-08-16 at 1 33 54 PM

## Notes
This PR relies on [Emma Lesson1 Red Phase](https://github.com/olioapps/typescript-starter-lab/pull/47)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,8 +9,12 @@ export const arrayCopier = (array: ReadonlyArray<People>) => {
 
 export const ageAdder = (people: ReadonlyArray<Partial<People>>, ages: Record<string, number>) => {
   const copy = [...people];
-  const addAges = copy.map((person, index) => {
-    return { ...person, age: ages[index]}
+  const addAges = copy.map((person) => {
+    const name = person.name?.toLowerCase();
+    if (name != null) {
+      return { ...person,
+         age: ages[name]}
+    }
   });
   return addAges;
 };
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,5 +56,10 @@ export const idAsName = (people: ReadonlyArray<People>) => {
 }
 
 export const justNames = (people: ReadonlyArray<People>) => {
-
+  const peopleCopy = arrayCopier(people);
+  const namesArray = peopleCopy.map(person => {
+    const { name } = person;
+    return name;
+  });
+  return namesArray;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,7 @@ export interface People {
 }
 
 export const arrayCopier = (array: ReadonlyArray<People>) => {
-
+  return [...array];
 };
 
 export const ageAdder = (people: ReadonlyArray<Partial<People>>, ages: ReadonlyArray<number>) => {
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,12 @@ export const arrayCopier = (array: ReadonlyArray<People>) => {
 };
 
 export const ageAdder = (people: ReadonlyArray<Partial<People>>, ages: ReadonlyArray<number>) => {
-
+  const copy = [...people];
+  const addAges = copy.map((person, index) => {
+    person.age = ages[index]
+  });
+  console.log(addAges)
+  return addAges;
 };
 
 export const ageRemover = (people: ReadonlyArray<People>) => {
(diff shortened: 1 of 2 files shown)


DuHerb:
PR note:
The `Relies On` section isn't one that we'd ever include in a PR.  I would add this type of information to `Additional Context / Notes` section.

src/index.ts
```diff
@@ -1,30 +1,63 @@
 export interface People {
```
DuHerb: Some TS notes:
i'd change `People` --> 'Person'
and then you can do `People = ReadonlyArray<Person>`

Question: why did you remove the readonly decorator from `name` and `age` props?
You don't need to make these changes now because I'm not concerned about TS for this excercise.

src/index.ts
```diff
@@ -1,30 +1,63 @@
...
 export const arrayCopier = (array: ReadonlyArray<People>) => {
-
+  return [...array];
 };
 
 export const ageAdder = (people: ReadonlyArray<Partial<People>>, ages: Record<string, number>) => {
-
+  const addAges = people.map((person) => {
+    const name: string | undefined = person.name?.toLowerCase();
+    if (name != null) {
```
DuHerb: be careful with explicit `null` checks.

if `name` is undefined, or other wise falsy, it may pass the conditional check.

here, `if (name) { ... }` would suffice.

src/index.ts
```diff
@@ -1,30 +1,63 @@
...
 
 export const arrayCopier = (array: ReadonlyArray<People>) => {
-
+  return [...array];
 };
 
 export const ageAdder = (people: ReadonlyArray<Partial<People>>, ages: Record<string, number>) => {
-
+  const addAges = people.map((person) => {
+    const name: string | undefined = person.name?.toLowerCase();
```
DuHerb: This level of complexity is not required for this exercise.

But if you want to include it, note that you are changing the string value of the `person.name` property, assuming that it's supposed to match up with a key in the 'Ages' object.

casing matters for Keys, so it is possible the `Aron` and `aron` aren't actually supposed to be related.

Typically, in a real application, there would be some unique identifier that you would use to relate pieces of data together, most commonly an `id`.

However, if there WAS a case whereyou are comparing non-cased characters, (ie: if you're doing a name search), then make sure that both the `search' string, and the value being checked are lowercased.

src/index.ts
```diff
@@ -1,30 +1,63 @@
...
+  });
+  return removeAges;
 }
 export const ageRemover2 = (people: ReadonlyArray<People>) => {
-
+  const removeAges = people.map(person => {
+    const { name } = person
+    return {
+      name: name
+    };
```
```suggestion
    return { name };
```

src/index.ts
```diff
@@ -1,30 +1,63 @@
...
 }
 
 export const idAsName = (people: ReadonlyArray<People>) => {
-
+  const replaceName = people.map(person => {
+    const { name: id, age } = person;
+    return {
+      id: id,
+      age: age
+    }
```
```suggestion
    return {
      id,
      age,
    }
```

src/index.ts
```diff
@@ -1,30 +1,63 @@
...
+  });
+  return replaceName;
 }
 
 export const justNames = (people: ReadonlyArray<People>) => {
-
+  const namesArray = people.map(person => {
+    const { name } = person;
+    return name;
+  });
```
DuHerb: Functionally, it doesn't make a difference and I can't imagine a use case where you'd need to alias a destructured key if you're just returning it's value. 

But a requirement of lesson 1.7 was that 
`// ALSO: use destructure aliasing to return the 'name' as 'id'`

I'm going to request the change here since it was an explicit part of the 'ticket'

DuHerb (changes requested):

src/index.ts
emmalgs: Fixed. I cannot remember why I didn't make name and age readonly...?

src/index.ts

src/index.ts
emmalgs: fixed! I was doing a weird work around for the types, but forgot that I could just use destructing to make a copy and call toLowerCase method on the copy in the computed property.

src/index.ts

src/index.ts

src/index.ts

PR #60, opened 2023-08-21
Emma init user api

### LINKED ISSUE #58: Emma - Init UserAPI
## What it is:
Initialize a `UserAPI` class object.

### Context:
The `UserAPI` object will manage its own collection of 'User' items.  `UserAPI` can be instantiated with a single, optional argument, which is an initial `User` repository.  `UserAPI` will contain expected CRUD methods, as well as some additional getter methods (to be defined in future tickets)

Acceptance Criteria:
- `User` object is typed, with the following properties
   - name
   - age
   - favColor
   - id (private)

- `UserAPI` has initial tests that verify that it can be instantiated with, or without, an initial 'User' collection
- `UserAPI` has a `getAllUsers` method that returns an array of all users in the repo
-  `UserAPI.getAllUsers` has a passing tests.

emmalgs:
## What this is:
Initial user api ticket
* Instantiate UserAPI class
* getAllUsers method

## Ticket
[init user api](https://github.com/olioapps/typescript-starter-lab/issues/58)

## Steps to verify
* in the project directory on the branch 'emma-init-user-api', run ```yarn test```

## Media
3 tests passing

## Notes
* I haven't tested maps before, so decided to compare the lengths of the incoming array to the size of the map? Was able to verify they are maps through console.logs. Not sure if this is typical and would appreciate feedback on best ways to test Map.


src/index.ts
```diff
@@ -1 +1,18 @@
-//Define class here
\ No newline at end of file
+export interface User {
+  readonly name: string,
+  readonly age: number,
+  readonly favColor: string,
```
DuHerb: We are sort of simulating a use case where we are using persisted data (eg: database records). these values MAY be updated in the future via an `updateUser` method.  So these should not be `readonly`.

But the instinct is generally correct, especially in regards to client side logic: you want to practice immutability, and enforce this pattern by defining types and their properties as ReadOnly unless it will explicitly need to be changed

src/index.ts
```diff
@@ -1 +1,18 @@
...
+  readonly id: string,
+}
+
+export class UserAPI {
+  users: ReadonlyMap<string, User>;
+  constructor(users: User[] = []) {
```
DuHerb: the constructor should consume a the actual shape of your collection. So, if the users collection is a map, then the UserAPI parameter should be a map of users.

The only thing the constructor might do is confirm that it has the correct shape.  However, I'm not concerned with that here.

DuHerb (changes requested):

src/index.ts
```diff
@@ -1 +1,18 @@
...
\ No newline at end of file
+export interface User {
+  readonly name: string,
+  readonly age: number,
+  readonly favColor: string,
+  readonly id: string,
+}
+
+export class UserAPI {
+  users: ReadonlyMap<string, User>;
```
DuHerb: please make `UserAPI.users` private, so that it is not accessable to any other actors.

This mirrors the way persisted data should be handled.  Only API's with explicit permission should be able to interact with data.

src/index.test.ts
```diff
@@ -1,7 +1,69 @@
...
+ } from ".";
 
-describe('Tests will go here!', () => {
-  it('should pass', () => {
-    expect(true).toBeTruthy()
+describe("UserAPI constructor", () => {
+  it("should instantiate an instance of UserAPI without user repository", () => {
+    const newUserApi = new UserAPI();
+    const expected = 0;
+    const actual = newUserApi.users.size;
```
DuHerb: confirm expected results by using the userApi.getUsers method, and other exposed methods.

src/index.ts

src/index.ts
emmalgs: addressed

src/index.ts

src/index.test.ts
emmalgs: fixed

src/index.ts
```diff
@@ -1 +1,17 @@
...
+export class UserAPI {
+  private users: Record<string, User>;
+  constructor(users: Record<string, User> = {}) {
+    this.users = users
+  }
+
+  getAllUsers() {
+    return [...Object.values(this.users)];
+  }
+}
```
DuHerb: new line at end of file.  this is a github lint rule

src/mockdata.ts
```diff
@@ -0,0 +1,22 @@
...
+  },
+  721: {
+    name: 'Opal',
+    age: 4,
+    favColor: 'teal',
+    id: "721"
+  }
+}
+
+export default userRepo;
```
DuHerb: new line at end of file

src/index.ts
DuHerb: these properties are still 'readonly'

DuHerb (changes requested):

src/index.ts

src/mockdata.ts
emmalgs: fixed

DuHerb (approved):

PR #61, opened 2023-08-21
Sarah init user api

### LINKED ISSUE #57: Sarah - Init UserAPI
## What it is:
Initialize a `UserAPI` class object.

### Context:
The `UserAPI` object will manage its own collection of 'User' items.  `UserAPI` can be instantiated with a single, optional argument, which is an initial `User` repository.  `UserAPI` will contain expected CRUD methods, as well as some additional getter methods (to be defined in future tickets)

Acceptance Criteria:
- `User` object is typed, with the following properties
   - name
   - age
   - favColor
   - id (private)

- `UserAPI` has initial tests that verify that it can be instantiated with, or without, an initial 'User' collection
- `UserAPI` has a `getAllUsers` method that returns an array of all users in the repo
-  `UserAPI.getAllUsers` has a passing tests.

reimannsarah:
**What this is**
Add User and UserAPI class, write to verify both classes can be instantiated
Write tests to verify UserAPI can be instantiated with and without an initial user repository passed as argument
Write getAllUsers method and passing tests

**Ticket**
[Init User Api](https://github.com/orgs/olioapps/projects/11?pane=issue&itemId=36441760)

**Steps to verify**
run `yarn test` in terminal

**Media**
Screenshot 2023-08-21 at 4 17 06 PM

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,28 +1,10 @@
-import { User, UserAPI } from './index'
-
-describe('User', () => {
-  it('should return a new user', () => {
-    const name = "Scorpo Fangoria";
-    const age = 16;
-    const favColor = "puce";
-    const id = 1;
-    const expected = {
-      name: "Scorpo Fangoria",
-      age: 16,
-      favColor: "puce",
-      id: 1
-    }
-
-    const actual = new User(name, age, favColor, id);
-
-    expect(actual).toEqual(expected);
-  })
-})
+import { UserAPI } from './index'
+import { users as user_repository } from './test_data';
 
 describe('UserAPI', () => {
   it('should create a new instance of UserAPI without taking any parameters', () => {
     const expected = {
-      users: []
+      users: {}
     }
 
     const actual = new UserAPI;
@@ -31,24 +13,21 @@ describe('UserAPI', () => {
   });
 
   it('should create a new instance of UserAPI and take a parameter of a user repository', () => {
-    const user1 = new User("Scorpo Fangoria", 16, "puce", 1);
-    const user2 = new User("Cadder Jenks", 67, "Barbie", 2);
-    const user_repository = [user1, user2];
     const expected = {
-      users: [
-        {
+      users: {
+        "243": {
           name: "Scorpo Fangoria",
           age: 16,
           favColor: "puce",
-          id: 1
+          id: 243
         },
-        {
+        "564": {
           name: "Cadder Jenks",
           age: 67,
           favColor: "Barbie",
-          id: 2
+          id: 564
         }
-      ]
+      }
     }
 
     const actual = new UserAPI(user_repository)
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1 +1,22 @@
...
+  constructor(initialUsers?: User[]) {
+    if(initialUsers) {
+      this.users = initialUsers;
+    }
+  }
+
+  getAllUsers(): User[] {
+    return this.users;
+  }
+}
```
DuHerb: new line at end of file

src/test_data.ts
```diff
@@ -0,0 +1,14 @@
...
+    favColor: "puce",
+    id: 243
+  },
+  "564": {
+    name: "Cadder Jenks",
+    age: 67,
+    favColor: "Barbie",
+    id: 564
+  }
+}
```
DuHerb, line 14: new line at end of file.  This is a github lint rule

src/index.test.ts
```diff
@@ -1,7 +1,62 @@
...
-    expect(true).toBeTruthy()
+describe('UserAPI', () => {
+  it('should create a new instance of UserAPI without taking any parameters', () => {
+    const expected = {
+      users: {}
+    }
+
+    const actual = new UserAPI;
+
+    expect(actual).toEqual(expected);
```
DuHerb: I don't think that this test is really asserting anything.

If you want to test that `actual` is an instance of `UserAPI`, then look at the `toBeInstanceOf` Jest assertion.

Note: you need to invoke the 'new` instance of UserAPI.  EG: `const actual = new UserAPI();`


If the point of the test is to make sure that initiating `UserAPI` with no argument creates an empty repo, then checking for the expected result of `UserAPI.getUsers()` is a better test.

Both of these would be valid tests.

src/index.test.ts
```diff
@@ -1,7 +1,62 @@
...
+          age: 67,
+          favColor: "Barbie",
+          id: 564
+        }
+      }
+    }
+
+    const actual = new UserAPI(user_repository)
+
+    expect(actual).toEqual(expected);
```
DuHerb: i'm not really sure why this test is passing, but I think the purpose is to make sure that it creates the `private users` object correctly.  I would remove this test,  since the `getAllUsers` test below accomplishes the desired outcome.

src/index.ts
```diff
@@ -1 +1,18 @@
...
+
+export class UserAPI {
+  private users: Record<string, User> = {};
+
+  constructor(initial_users: Record<string, User> = {}) {
+    this.users = initial_users;
+  }
+
+  getAllUsers(): Record<string, User> {
+    return this.users;
```
DuHerb: this function should return an array of users.

src/index.test.ts
```diff
@@ -1,7 +1,62 @@
...
+      },
+      "564": {
+        name: "Cadder Jenks",
+        age: 67,
+        favColor: "Barbie",
+        id: 564
+      }
+    }
+
+    const actual = userAPI.getAllUsers();
```
DuHerb, line 31: refactor this test per the comment on `index.ts#getAllUsers`

DuHerb (changes requested):

src/index.ts
```diff
@@ -1 +1,19 @@
-//Define class here
\ No newline at end of file
+type User =  {
+  name: string, 
+  age: number,
+  favColor: string,
+  id: number
```
DuHerb: `id` should be readonly.  In general, IDs will always be unique and should not be changed.

PR #64, opened 2023-08-22
Emma user api crud

### LINKED ISSUE #62: Emma - UserAPI CRUD
## What it is:
Complete full CRUD functionality on UserAPI class.

### Context:
The `UserAPI` object will manage its own collection of 'User' items.  `UserAPI` can be instantiated with a single, optional argument, which is an initial `User` repository.  `UserAPI` will contain expected CRUD methods, as well as some additional getter methods. These further methods include a getUserById, addUser, and deleteUserById.

Acceptance Criteria:
- `UserAPI` has (almost) full CRUD functionality (Create - addUser, Read - getAllUsers, getUserById, Update, and Delete - deleteUserById)
- `UserAPI` has three new methods: getUserById, addUser, and deleteUserById
- `UserAPI.getUserById` takes an id string and returns a single User object.
- `UserAPI.addUser` takes an uninitialized user object and creates a unique id, and adds the user to the repo with this id as its key.
- `UserAPI.deleteUserById` takes an id string and removes the user object
- All new methods have passing tests

emmalgs:
## What this is:
UserAPI CRUD
* complete CRUD methods for UserAPI class
* include addUser, deleteUserById, getUserById

## Ticket
[UserAPI CRUD](https://github.com/olioapps/typescript-starter-lab/issues/62)

## Steps to verify
* in the project directory on the branch 'emma-user-api-crud', run `yarn test`

## Media
10 tests passing

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -73,18 +73,18 @@ describe("UserAPI.deleteUserById", () => {
   it("should remove a user from a larger repo when given an id string", () => {
     const newUserApi = new UserAPI(userRepo);
     const expected = "721";
-    const actual = newUserApi.deleteUserAtId("721").id;
+    const actual = newUserApi.deleteUserById("721").id;
     expect(actual).toEqual(expected);
   });
   it("should return the deleted user", () => {
     const newUserApi = new UserAPI(userRepo);
     const expected = newUserApi.getUserById("342");
-    const actual = newUserApi.deleteUserAtId("342");
+    const actual = newUserApi.deleteUserById("342");
     expect(actual).toEqual(expected);
   });
   it("should throw an error if invalid user id was passed", () => {
     const newUserApi = new UserAPI(userRepo);
-    const error = () => newUserApi.deleteUserAtId("");
+    const error = () => newUserApi.deleteUserById("");
     expect(error).toThrow("Unable to delete user");
   });
 });
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -34,12 +34,18 @@ describe("UserAPI.addUser", () => {
     const actual = newUserApi.getAllUsers().length;
     expect(actual).toEqual(expected);
   });
-
   it("should generate an id for a new user when added to the UserAPI repo", () => {
     const newUserApi = new UserAPI();
     newUserApi.addUser(newUser);
     expect(newUserApi.getAllUsers()[0].id).toBeTruthy();
   });
+  it("should append new user to existing repo", () => {
+    const newUserApi = new UserAPI(userRepo);
+    newUserApi.addUser(newUser);
+    const expected = 4;
+    const actual = newUserApi.getAllUsers().length;
+    expect(actual).toEqual(expected);
+  })
 });
 
 describe("UserAPI.getUserById", () => {
@@ -55,6 +61,11 @@ describe("UserAPI.getUserById", () => {
     newUserApi.addUser(newUser);
     const actual = newUserApi.getUserById("2");
     expect(actual).toBeFalsy();
-  })
+  });
 });
 
+describe("UserAPI.deleteUserById", () => {
+  it("should remove user at a given id string", () => {
+
+  })
+})
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -50,10 +50,9 @@ describe("UserAPI.addUser", () => {
 
 describe("UserAPI.getUserById", () => {
   it("should return a single user object when matching id string is passed as an argument", () => {
-    const newUserApi = new UserAPI();
-    newUserApi.addUser(newUser);
-    const expected = { ...newUser, id: "1" };
-    const actual = newUserApi.getUserById("1");
+    const newUserApi = new UserAPI(userRepo);
+    const expected = newUserApi.getAllUsers()[0];
+    const actual = newUserApi.getUserById("342");
     expect(actual).toEqual(expected);
   });
   it("should return falsy if no id matches a user", () => {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1,17 +1,39 @@
...
+
+  getUserById(id: string) {
+    return this.getAllUsers().filter((user) => user.id == id)[0];
+  }
+
+  deleteUserAtId(id: string) {
+    delete this.users[id];
+  }
+
+  assignId() {
```
DuHerb: Having 'helper' functions like `assignId` is great.  But these helpers don't need to be exposed.  So it should be marked as private.

This is one way to guard access to data and functionality.

DuHerb: Naming nit pick.  `assignId` => `getUniqueId` or `generateUid`. Something that describes what the function is actually doing.

src/index.ts
```diff
@@ -1,17 +1,39 @@
...
+    this.currentId += 1;
```
DuHerb: `this.currentId +=1` is a reasonable first approach to providing a unique id/key.

But think about cases where `UserAPI` is instantiated with a set of users.  Those users already have ID's, and there are no checks here to guarantee that the new ID is actually unique.

There are many ways to generate UUIDs, but for our use case here, the simplist approach is to use a stringified timestamp. Try using the javascript `Date` api to generate the current time as a string, and use that as the id.

src/index.ts
```diff
@@ -1,17 +1,39 @@
...
+  addUser(user: User) {
+    const newId = this.assignId();
+    this.users[newId] = {
+      ...user,
+      id: newId.toString(),
+    };
+  }
+
+  getUserById(id: string) {
+    return this.getAllUsers().filter((user) => user.id == id)[0];
```
DuHerb: you should be able to access `this.users` directly, using the `id` as the lookup value.

src/index.ts
```diff
@@ -1,17 +1,39 @@
...
+  }
+
+  deleteUserAtId(id: string) {
+    delete this.users[id];
```
DuHerb: I'm making a new request here.  Please return the deleted user object.  This is a common pattern that you'll run into.  Often, an application will want to do something with some item that was just deleted -- IE: have the UI display information about that item.

DuHerb: Secondly, i would like you to use an immutable solution to this.  EG: don't operate directly on the `this.users` object.  using `delete` and `add` are probably very safe to use in this `UserAPI`.  But i'd like you to practice using immutable techniques.  use destructuring to remove the 'toBeDeletedUser'

EG: copy the existing state.  Transform/handle the copied state.  Return the handled copy as the new state.

DuHerb (changes requested):

src/index.ts

src/index.ts
emmalgs: fixed -- needed to adjust some tests with this fix

src/index.ts
emmalgs: fixed!

src/index.ts

src/mockdata.ts
```diff
@@ -19,4 +19,10 @@ const userRepo = {
...
 
-export default userRepo;
\ No newline at end of file
+const newUser = {
+  name: 'Pete',
+  age: 1,
+  favColor: 'blue',
+}
+
+export { userRepo, newUser };
```
DuHerb: new line at end of file.  We should be reviewing our PRs to catch lint errors like these before submitting for review.

src/index.ts
```diff
@@ -1,17 +1,44 @@
...
   }
 
-  getAllUsers() {
+  getAllUsers(): ReadonlyArray<IdAwareUser> {
     return [...Object.values(this.users)];
   }
-}
\ No newline at end of file
+
+  addUser(user: User): void {
```
DuHerb: many `create` type methods will require that the newly added object is returned.  The requesting client will likely wish to do something with the result.  IE: the client may need the new `id` to create some other resource.

go ahead and return the new object, and update your test so that it confirms that the new object has the expected properties.  name == expected name, id exists, etc.

src/index.test.ts
```diff
@@ -25,3 +25,56 @@ describe("UserAPI.getAllUsers", () => {
...
+  });
+});
+
+describe("UserAPI.deleteUserById", () => {
+  it("should remove a user from a larger repo when given an id string", () => {
+    const newUserApi = new UserAPI(userRepo);
+    const expected = newUserApi.getAllUsers().length - 1;
+    newUserApi.deleteUserAtId("721");
+    const actual = newUserApi.getAllUsers().length;
+    expect(actual).toEqual(expected);
```
DuHerb, line 77: confirm that the correct user was deleted by inspected the return properties.  Does it have the correct id?

DuHerb (changes requested):

src/mockdata.ts
emmalgs: fixed

src/index.ts

src/index.test.ts
emmalgs: fixed

src/index.ts
```diff
@@ -1,17 +1,56 @@
...
+
+  getUserById(id: string): IdAwareUser {
+    if (!this.users[id]) {
+      throw new Error('User not found')
+    } else {
+      return {...this.users[id]};
+    }
+  }
+
+  deleteUserAtId(id: string): IdAwareUser {
```
DuHerb: `deleteUserById` is Olio convention for these types of functions.  `getUserById` `updateUserById` etc

DuHerb (approved):

src/index.ts

PR #65, opened 2023-08-22
Sarah user api crud

### LINKED ISSUE #63: Sarah - UserAPI CRUD
## What this is
Complete Create, Read, and Delete functionality for `UserAPI`

## Context
`UserAPI` currently has Read functionality with `getAllUsers`, need to finish the rest of the CRUD functions.

## Acceptance Criteria
- `UserAPI` has `getUserByID` function
     - takes id as param 
     - returns single object
- `addUser` function
     - takes initial user w/o id as a param
     - return: creates a unique id for the user and adds the user + id to user repo
- `deleteUserById`
     - takes id as param
     - removes user from object
 -all methods have passing tests

reimannsarah:
**What this is**
Add rest of CRUD methods:
`addUser`
`getUserById`
`deleteUserById`
add passing tests for all methods

**Note**
This PR relies on [Sarah Init User Api](https://github.com/olioapps/typescript-starter-lab/pull/61)

**Ticket**
[Sarah - UserAPI CRUD](https://github.com/orgs/olioapps/projects/11/views/1?pane=issue&itemId=36551302)

**Steps to verify**
run `yarn test` in terminal 

**Media**
Screenshot 2023-08-25 at 11 16 51 AM


src/index.ts
```diff
@@ -12,8 +12,35 @@ export class UserAPI {
...
-    const users_array = Object.values(this.users);
-    return users_array;
+  getAllUsers(): Record<string, User> {
+    return this.users;
   }
-}
+
+  addUser(userObj: User) {
+    const newUserId = (Math.floor(Math.random() * 900)).toString();
+    // const newUserId = v4();
```
DuHerb: remove dev comments

DuHerb: I see why you pulled out `math.random` for this problem.  However, `random` doesn't guarantee uniqueness.

Try using a 'timestamp' based solution.

If you wanted/needed to preserve the first approach, an additional safeguard would be check and see if a generated ID already exists in your repo before using it. If it doesn't, make another one.  NOTE: That solution involves recursion and goes beyond the scope of this exercise.  Only try it if you are waiting for work.

src/index.ts
```diff
@@ -12,8 +12,35 @@ export class UserAPI {
...
-  getAllUsers(): ReadonlyArray<User> {
-    const users_array = Object.values(this.users);
-    return users_array;
+  getAllUsers(): Record<string, User> {
+    return this.users;
   }
-}
+
+  addUser(userObj: User) {
+    const newUserId = (Math.floor(Math.random() * 900)).toString();
```
DuHerb: please create a private method that returns a unique id.  

This helps build 'composable' functions, as well as practices the 'separation of concerns'.  `addUser`s concern is adding a new user to the existing user collection.  `getUid` will be concerned with generated a unique id.

`addUser` can invoke `getUid` as part of it's body, but each function remains testable and maintainable independently of each other.

src/index.ts
```diff
@@ -12,8 +12,35 @@ export class UserAPI {
...
+      this.users[id]
+    )
+  }
+
+  deleteUserById(id: string) {
+    delete this.users[id];
+    if(this.users[id]) {
+      return this.users[id]
+    } else {
+      return null
```
DuHerb: nice catch on handling the 'user does not exist' condition.  The upcoming error handling ticket will update this function to throw an error, rather than return null.

DuHerb (changes requested):

src/index.ts
```diff
@@ -2,18 +2,56 @@ type User =  {
   name: string, 
   age: number,
   favColor: string,
-  readonly id: number
+}
+
+type IUser = {
```
DuHerb: IUser should `extend` user

src/index.test.ts
```diff
@@ -33,3 +33,54 @@ describe('getAllUsers', () => {
...
 })
 })
+
+describe('addUser', () => {
+  it('should create a unique id for an inputted user object and assign the id to the key of the user object in the users map', () => {
+    const userApi = new UserAPI();
+    const expected = userApi.getAllUsers().length + 1;
+    userApi.addUser(new_user);
+
+    const actual = userApi.getAllUsers().length;
```
DuHerb: another assertion here would be to confirm that the new user actually exists, as expected.  You could find the `new_user` object in the `actual` value, and confirm that expected property values exists.  ie: new_user.name === added_user.name`

src/index.ts
```diff
@@ -1,19 +1,60 @@
...
+    const user_array = Object.values(this.users);
+    return user_array;
+  }
+
+  addUser(userObj: User): string {
+    const newUserArr = { ...this.users }
+    const id_user = this.assignUniqueId(userObj);
+    newUserArr[id_user.id] = id_user;
+    this.users = newUserArr;
+    return "User successfully added";
```
DuHerb: the instinct to return some confirmation is good.  If `addUser` was an http request, you'd expect a 200 and, possibly, a payload to be returned.

a common pattern here for a `create` method would be to return the new, added user.

The created object may have new data attached  to it that the requesting client may want to work with.  IE: user.id.

I would refactor this function to return the newly created user object.

src/index.ts
```diff
@@ -1,19 +1,60 @@
...
+    return (
+      this.users[id]
+    )
+  }
+
+  deleteUserById(id: string): string {
+    const newUserArr = { ...this.users };
+    delete newUserArr[id];
+    this.users = newUserArr;
+    return "User successfully deleted";
```
DuHerb: same idea as `addUser`.  return the user object that was just deleted.

src/index.test.ts
```diff
@@ -33,3 +33,54 @@ describe('getAllUsers', () => {
...
+    const userApi = new UserAPI(user_repository);
+    const id = "564"
+    const expected = {
+      name: "Cadder Jenks",
+      age: 67,
+      favColor: "Barbie",
+      id: "564"
+    }
+
+    const actual = userApi.getUserById(id);
```
DuHerb:
```suggestion
    const actual = userApi.getUserById(expected.id);
```



use `expected.id` rather than `const id`.  Couple of reasons: less code, and it ensures that your code is using the correct data or values.

if your test involved testing mismatched id's, then `const fooID` would be necessary.

this is a super nitpick comment.  :)

src/index.test.ts
```diff
@@ -33,3 +33,54 @@ describe('getAllUsers', () => {
...
+})
+
+describe('deleteUserById', () => {
+  it('should remove the user at the inputted id from the users object', () => {
+    const userAPI = new UserAPI(user_repository);
+    const id = "564";
+    const expected = userAPI.getAllUsers().length -1;
+
+    userAPI.deleteUserById(id);
+    const actual = userAPI.getAllUsers().length;
```
DuHerb: add assertion that the deleted user is actually the expected object.  (see comment on `index.ts#deleteUserById`)

DuHerb (changes requested):

src/index.test.ts
```diff
@@ -33,3 +33,51 @@ describe('getAllUsers', () => {
     expect(actual).toEqual(expected);
 })
 })
+
+describe('addUser', () => {
+  it('should create a unique id for an inputted user object and assign the id to the key of the user object in the users map', () => {
```
DuHerb: the description of this test is a little long.  shorten it up.
we are describing add user.
if this test is verfiying that new users have Ids, then the test should check for an id.


checking that the name, or other known properties is also ok for a test like this.  Really, this test is about verifying that a new user was successfully created and returned by `UserAPI.addUser`.  so verifying that the returned user has the expected fields is appropriate to do in a single test.

So we could call this test `should return expected IDAwareUser`

src/index.test.ts
```diff
@@ -33,3 +33,51 @@ describe('getAllUsers', () => {
...
+    const userApi = new UserAPI();
+
+    const actual =  userApi.addUser(new_user);
+
+    expect(actual.name).toEqual(new_user.name);
+  })
+
+  it('should return the added user if it is added successfully', () => {
```
DuHerb: this test is redundant.  the above test is already covering the fact that a new user has been added successfully.

src/index.test.ts
```diff
@@ -33,3 +33,51 @@ describe('getAllUsers', () => {
...
+    const userApi = new UserAPI();
+
+    const actual = userApi.addUser(new_user);
+
+    expect(actual.id).toBeTruthy;
+  })
+})
+
+describe('getUserById', () => {
+  it('should take an id as a parameter and return a single object', () => {
```
DuHerb:
```suggestion
  it('should return expected IDAwareUser', () => {
```

src/index.test.ts
DuHerb: Checking the length of the array after `deleteUserByID` is called is only partially useful.  Please verify that the returned user (the one that was deleted), is the expected user.

DuHerb (changes requested):

src/index.ts
reimannsarah: addressed

src/index.ts
reimannsarah: addressed

src/index.ts
reimannsarah: addressed

src/index.ts
reimannsarah: addressed

src/index.test.ts
reimannsarah: addressed

src/index.ts
reimannsarah: addressed

src/index.test.ts
reimannsarah: addressed

src/index.ts
reimannsarah: addressed

src/index.test.ts
reimannsarah: addressed

src/index.test.ts
reimannsarah: addressed

src/index.test.ts
reimannsarah: addressed

src/index.test.ts
reimannsarah: addressed

PR #75, opened 2023-08-24
EventStream Puzzler 

reimannsarah:
**What this is**
Red phase tests for eventStream puzzler

**Ticket**
[Sarah-EventStream-Puzzler](https://github.com/orgs/olioapps/projects/11/views/1?pane=issue&itemId=36774618)

**Steps to Verify**
run `yarn test` in terminal

**Media**
Screenshot 2023-08-25 at 4 29 38 PM


src/index.ts
```diff
@@ -1 +1,16 @@
...
+}
+
+type highest_score = {
+  events: ReadonlyArray<event>,
+  score: number
+}
+
+export function scoreEventStream(eventStream: ReadonlyArray<event>): highest_score {
+  return {
+    events: [],
```
reimannsarah: I added a return object here because otherwise my test suite would not run and I have typed the output for the function

src/index.test.ts
```diff
@@ -1,7 +1,51 @@
...
+    const actual = scoreEventStream(eventStream);
+
+    expect(actual).toHaveProperty("events");
+  })
+
+  it('should return an object with a score property', () => {
+    const actual = scoreEventStream(eventStream);
+
+    expect(actual).toHaveProperty("score");
+  })
```
DuHerb: these two tests can be combined, since you're testing that the return object has the expected properties.

just do both expects in the same test.

src/index.test.ts
```diff
@@ -1,7 +1,51 @@
...
+
+  it('successfully assess the length of the inputted array', () => {
```
DuHerb: future tests should include checks for different region lengths?  If not, just add a comment here saying that the default length is 5.

src/index.ts
```diff
@@ -1 +1,16 @@
-//Define class/functions here
+type event = {
```
DuHerb: interface

src/index.ts
```diff
@@ -1 +1,16 @@
...
+  timestamp: number,
+  eventType: string
+}
+
+type highest_score = {
```
DuHerb: interface

DuHerb: high_score -> HighScore

src/index.test.ts
```diff
@@ -1,7 +1,32 @@
...
-    expect(true).toBeTruthy()
+describe('scoreEventStream', () => {
+  it('should return an object with an events property and a score property', () => {
+    const actual = scoreEventStream(eventStream);
+
+    expect(actual).toHaveProperty("events");
+    expect(actual).toHaveProperty("score");
+  })
+
+  it('length of events array in return object should be no greater than 5', () => {
```
DuHerb, line 12: since the function now accepts the regionLength argument, then we need additional tests to verify expected array lengths.  Default vs > vs <

src/models.ts
```diff
@@ -0,0 +1,13 @@
+export interface event {
```
DuHerb, line 1:
```suggestion
export interface Event {
```

DuHerb: interfaces and types should be capialized.

DuHerb:
```
type EventStream = ReadonlyArray<Event>;
```

src/index.ts
```diff
@@ -1 +1,72 @@
...
+    container.push(event);
+  }
+  return container;
+}
+
+function createScoresArray(eventStreamWithScores: ReadonlyArray<scored_region> | ReadonlyArray<eventWithScore>) {
+  return eventStreamWithScores.map(e => { return e.score })
+}
+
+export function scoreEventStream(eventStream: ReadonlyArray<event>, subRegionLength: number = 5) {
```
DuHerb, line 54: `const DEFAULT_SUBREGION_LENGTH = 5`

PR #77, opened 2023-08-25
Emma eventstream puzzler pass tests

### LINKED ISSUE #76: Emma - EventStream Puzzler
### What it is:
Implementation of eventstream algorithm

### Context:
A list of events (with a timestamp and an event type) is passed to a function, eventScorer, along with the number of events that should be consecutively scored. The expected output of this function is the highest scored consecutive events. The scoring criteria is as follows: messages are worth 1, views are worth 2, and screenshots are worth 3. The goal is to find the highest scoring consecutive events in a given region.

### Acceptance Criteria
* The function should take 2 parameters: an array of Events and a subregion length
* The output should be an object of the highest scoring region with the score of that region
* All tests should be passing
* The default region length should be 5

emmalgs:
### What It Is:
Solves eventStream puzzler
* passes all tests

### Ticket
[Emma EventStream Puzzler](https://github.com/orgs/olioapps/projects/11/views/1?pane=issue&itemId=36795651)

### Steps to Verify
In the project directory on the branch `emma-eventstream-puzzler-pass-tests` run `yarn test`

### Media
passing event puzzler tests

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,1 +1,15 @@
-//Define class/functions here
+type EventObject = {
+  timestamp: number,
+  eventType: string,
+}
+
+type Score = {
+  subregion: Array<EventObject>,
+  score: number,
+}
+
+const eventStream = (events: Array<EventObject) => {
+
+}
+
+export { eventStream, EventObject };
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,7 +40,7 @@ const scoreEventStream = (events: Array<Event>, regionLength: number = 5) => {
     for (let i = 0; i < events.length - regionLength; i++) {
       const subregion = events.slice(i, i + regionLength);
       const score = addScore(subregion);
-      if (score >= result.score || !result.score) {
+      if (score >= result.score) {
         (result.events = subregion), (result.score = score);
       }
     }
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -8,16 +8,44 @@ type Score = {
...
+  } else {
+    const result: Score = {
+      events: [],
+      score: 0,
+    };
+    for (let i = 0; i < events.length - regionLength; i++) {
+      const subregion = events.slice(i, i + regionLength);
+      const score = addScore(subregion);
+      if (score >= result.score) {
+        (result.events = subregion), (result.score = score);
```
```suggestion
        result.events = subregion;
        result.score = score;
```

src/index.ts
```diff
@@ -8,16 +8,44 @@ type Score = {
   score: number;
 };
 
-// rename to scoreEventStream
+const addScore = (subregion: Array<Event>) => {
```
DuHerb: see if you can break this down even more.

It would be useful, and more modular to have a function that gets the score for a single event.

That function can be used in another helper function that gets the total score from an array of events

src/index.ts

src/index.ts

PR #118, opened 2023-08-31
Add a single passing test

reimannsarah:
**What this is**
Add a passing test for readAllUserDataFiles

**Context**
It took a very long time to figure out how to test the FileSystem code that co-pilot wrote but ultimately we went with a beforeEach and afterEach solution to create some test data and then tear it down after the test runs. We used chatGPT a lot to help problem solve the testing. It sent us down a rabbit hole of creating mock functions which we ultimately did not end up using. 

**Steps to verify**
run `yarn test` in terminal

**Media**
Screenshot 2023-08-31 at 4 24 31 PM


emmalgs (approved):
Looks great to me!

PR #128, opened 2024-04-03
Classes and Types for EventStream Puzzler

KameronKeller:
# What this is:
* Implement classes that are expected to be used for implementing the algorithm
* `models/highestregionlocator.ts` defines a class that receives an event stream and will have a method that can locate the highest region in the dataset
  * The `locateHighestRegion()` function exists as a placeholder and will throw an error if called
* `models/region.ts` defines a class that can add up the scores of the app events that are added to it
* `index.ts` now has a main function and a console.log as a placeholder to verify it is being called when the program is being run
* `types.ts` defines types for AppEvents, EventStream data, and the dictionary for the Scores class
* `index.test.ts` defines high-level tests for validating that the classes created above successfully interface with each other

# Issue:
N/A

# Media:

https://github.com/olioapps/typescript-starter-lab/assets/17558364/d45a6cc5-1ea2-4e8b-89e6-cfa20d323243

# Steps to Verify: 

* Clone the branch
* Run `yarn install`
* Run `yarn test`


src/models/eventstream.ts
```diff
@@ -0,0 +1,10 @@
+import { EventStreamInterface } from "../types";
+import { AppEvent } from "./appevent";
+
+export class EventStream {
+    events: AppEvent[];
```
KameronKeller: I am curious if this is an appropriate way to assign a type to a variable. AppEvent is a class - would it be better to define an interface or type that defines the type that events should be?

src/assets/sampleData.ts
```diff
@@ -0,0 +1,38 @@
+const data = [
```
rachoac: type your sample data

src/index.test.ts
```diff
@@ -1,7 +1,94 @@
-//write tests here
+import { AppEvent } from "./models/appevent"
+import { EventStream } from "./models/eventstream"
+import { HighestRegionLocator } from "./models/highestregionlocator"
+import { Region } from "./models/region";
```
rachoac: no need for semicolons

src/index.test.ts
```diff
@@ -1,7 +1,94 @@
...
+      timestamp: 123123125,
+      eventType: "newMessage",
+    },
+    {
+      timestamp: 123123125,
+      eventType: "newMessage",
+    }
+  ]
+
+  let eventStream = new EventStream(sampleData);
```
rachoac: if not ever redefining eventStream, use const

src/index.test.ts
```diff
@@ -1,7 +1,94 @@
...
+
+  it('instantiates an EventStream object', () => {
```
rachoac: use ' and " consistently throughout your project

src/index.test.ts
```diff
@@ -1,7 +1,94 @@
...
+  let appEvent = new AppEvent(
+    {
+      timestamp: 123123123,
+      eventType: "newMessage",
+    }
+  );
+
+  
+  it('instantiates a Region object', () => {
+    let region = new Region();
```
rachoac: no need to test the construction of objects

src/index.ts
```diff
@@ -1 +1,7 @@
-//Define class here
\ No newline at end of file
+//Define class here
+
+function main() {
```
rachoac: use es6 

const main = () => {
...
}

src/models/appevent.ts
```diff
@@ -0,0 +1,14 @@
+import { AppEventInterface } from "../types"
+import { Scores } from "./scores";
+
+export class AppEvent {
+    timestamp: number;
```
rachoac: are these meant to be publicaly accessible? should they be private?

rachoac: why is AppEvent a class? is it going to have methods?
if not then this is better off as an interface or type

src/models/scores.ts
```diff
@@ -0,0 +1,13 @@
+import { ScoresDictionary } from "../types";
+
+export class Scores {
```
rachoac: maybe we don't need Scores to be a class, since its not stateful (has no no member variables). this might better be a collection of exported utility functions?

src/types.ts
```diff
@@ -0,0 +1,10 @@
+export interface AppEventInterface {
```
rachoac: what is the difference between interface and type?

why would you use one versus another, in typescript? based on this, does it make sense to define these as types, or interfaces?

src/index.test.ts
KameronKeller: Semicolons removed

src/index.test.ts
KameronKeller: Updated variables to const when able

src/index.test.ts
KameronKeller: Replaced all double quotes with single quotes

src/index.ts
KameronKeller: Switched to es6 format

src/models/scores.ts
KameronKeller: Relocated this functionality to a utility functions file and removed this class

src/models/appevent.ts
KameronKeller: Refactored as a type

src/index.test.ts
KameronKeller: Test removed

src/assets/sampleData.ts
KameronKeller: Type added

src/models/highestregionlocator.ts
```diff
@@ -0,0 +1,16 @@
+// import { EventStream } from './eventstream'
+import { Region } from './region'
+import { EventStream } from '../types'
+
+export class HighestRegionLocator {
```
rachoac: camelCase the file

src/index.test.ts
```diff
@@ -1,7 +1,91 @@
-//write tests here
+// import { AppEvent } from './models/appevent'
```
rachoac: avoid dead code

src/index.test.ts
```diff
@@ -1,7 +1,91 @@
...
+    {
+      timestamp: 123123125,
+      eventType: 'newMessage',
+    }
+  ]
+
+  const eventStream: EventStream = sampleData
+
+  it ('instantiates a HighestRegionLocator object', () => {
+    const highestRegionLocator = new HighestRegionLocator(eventStream)
```
rachoac: no need to test instance

src/models/region.ts
```diff
@@ -0,0 +1,19 @@
+import { retrieveScoreByEventType } from '../utils/utils'
+import { AppEvent } from '../types'
+
+export class Region {
+    regionEvents: AppEvent[] = []
+    score: number = 0
```
rachoac: do we need this?

src/models/highestregionlocator.ts
KameronKeller: Changed filename to camel case

src/index.test.ts
KameronKeller: Removed unneeded instance test

src/models/region.ts
KameronKeller: Removed score member variable as it was not being used and won't be needed

src/index.test.ts
KameronKeller: Dead code removed

src/types.ts
KameronKeller: Switched to types, primarily because interfaces can be "merged" together which could be unexpected

